PR #3, opened 2017-05-09
Bugfix: Player can be pushed over chains

Hypexion:
Player could be pushed over diagonal Gentryii chains.
See thread: http://forum.caravelgames.com/viewtopic.php?TopicID=41237

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,16 @@
+# Dependancies
+Deps/
+DepsSrc/
+
+# Build Folders
+**/Release/
+**/Release-Temp/
+**/Debug/
+**/Debug-Temp/
+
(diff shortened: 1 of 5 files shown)


RetrocadeNet:
@Hypexion I've merged the PR. Could you setup your editor to use tabs instead of spaces? As far as I can tell we mostly use tabs (except the few places where spaces sneaked in) for indents. For this one let's keep it as-is but would be great if the future ones had correct formatting :)

RetrocadeNet:
Merge commit: https://github.com/RetrocadeNet/drod/commit/f06b5c6b1edc888fc2a02a9761c01a6c0ba5a56e

PR #4, opened 2017-05-16
Fix single step

IQubic:
This is a fix for the Single Step bug as pointed as Nuntar here:
http://forum.caravelgames.com/viewtopic.php?TopicID=42309&page=0#416991

I also noticed that move (relative version) waited for a single turn too long after a failed turn. I went ahead and fixed that too.


IQubic:
Retrocade, how should I remove those?

PR #5, opened 2017-05-16
Fix for both types of Movement

IQubic:
Fix for Move Command and Single Step.
http://forum.caravelgames.com/viewtopic.php?TopicID=42309

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ DepsSrc/
 **/*.dll
 **/*.lib
 **/*.exe
+data/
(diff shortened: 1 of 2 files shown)


RetrocadeNet:
@IQubic  Thanks for the PR, unfortunately there are a couple of problems I'll have to ask you to fix:
 1. Can you revert your `.gitignore` changes? I've pushed a new version of the file and it should cover all the files that have to be ignored.
 2. Unfortunately we can't change the current behavior, please check the thread for more information about what to do here: http://forum.caravelgames.com/viewtopic.php?TopicID=42309&page=0#417126

PR #10, opened 2017-07-02
Allow Slayers/Guards to body kill other monsters when armed with dagger.

Hypexion:
As a resolution to these threads:

http://forum.caravelgames.com/viewtopic.php?TopicID=40922
http://forum.caravelgames.com/viewtopic.php?TopicID=40933

both Slayers and Guards have been given the ability to body kill monsters if they are armed will a dagger. As with sword attacks, both monsters will not body kill Guards or Slayers, and continue to treat them as movement obstacles.


RetrocadeNet:
@Hypexion My spider sense is telling me there are no tests for this task. Do you think it would be reasonable to add a few cases that prove it works?

PR #14, opened 2018-03-11
Correct monster names

Hypexion:
What if we lived in a world where the Aumtlich and Waterskipper source files were correctly named?

This pull request makes that dream reality. It also updates the MIDs for the two monsters, making everything flawlessly consistent.


IQubic:
If we are going to updating monster names, will we also be changing Splitter to Rock Giant? What about Ghost to Seep, and Phoenix to Fegundo? The seep and fegundo name changes aren't as important because you can tell what those are. I don't think it is obvious what Splitter is just from the name.

PR #21, opened 2020-09-06
Update url and filepaths for metakit component

Hypexion:
Sometime in the past, the metakit people took down their Github and moved somewhere else. This means that the dependencies script can't download it.

I've updated the url to the new one for `metakit 2.4.9.5`. It also seems to be the case that the filepaths are now different when the archive is unzipped, so I've updated those as well.

All other dependencies are working as of `05/09/2020`


Hypexion:
It looks like this is no longer needed with the latest script updates for VS2019.

mrimer:
Thanks for taking note of this issue!  Sounds like you're okay with moving to the available 2.4.9.7 version.

PR #20, opened 2019-05-01
Set Behavior script command

Hypexion:
Adds a new script command to allow great control over how NPCs interact with the room. Based on this thread: http://forum.caravelgames.com/viewtopic.php?TopicID=43392

# Implementation Details
To store an NPC's behaviours, a new sub-set of script flags was added. These are stored in a standard C++ set member variable in `CCharacter`. If a behavior's value is in the set, the behaviour is considered to be "On", otherwise it is considered to be "Off". Active behaviour alter how NPCs interact with various objects. When a behaviour check is needed externally, a public method has been added to the `CCharacter` header.

To maintain backwards compatibility, some behaviours are set by default in `CCharater::SetCurrentGame()`

# Specific Behaviour Details
Many behaviors required changes to both NPC and other element's functionallity. These changes are listed below, in three broad catagories of "behaviour types".

## Room Interaction
These behaviours affect how the NPC interacts with various room elements, such as trapdoors and tokens. These behaviours only required changes to `CCharacter` to be functional.

**Activates Tokens** replaces the IsHuman check with a behaviour check. By default, human NPCs can still activate tokens.
Trapdoor interaction behaviours change the checks in `CCharacter::CanDropTrapdoor()` from being based on the NPC's indentity to behaviour checks. Player double and Construct NPCs gain the appropriate behaviours to retain their previous interaction.
**Pushes Objects** and **Move Platforms** both required the removal of human specific logic in `CCharacter::DoesSquareContainObstacle()`. Rather than assuming that the NPC is able to push things, checks in `CCharacter::DoesSquareContainObstacle()` now check behaviours. Human NPCs gain these behaviours to retain previous interactions.

These changes together removal the need for human-specific logic in `CCharacter::MoveCharacter()`, as interactions are now controlled by behaviors.

## Monster Targetting
Five behaviours exist in this catagory, but only two are implemented. The non-implemented behaviours require signifigant refactoring to be functional.

**Attackable by Monsters** requires a change to `CCharacter::IsAttackableTarget()` to use a behaviour check, rather than using the NPC's identity. Smitemaster and Stalwart NPCs gain these behaviours to retain previous interactions.
**Puff Target** requires a change to `CFluffBaby::GetGoal()`. The function now looks at a monster's type, rather than its resolved identity. This allows a behaviour check to made when an NPC is a potenial target. If the NPC is not a Puff Target, the function will move onto the next monster.

## Damage Immunity
[...]

PR #6, opened 2017-05-25
Wait for weapon

Hypexion:
Pull request for script command 'Wait for Weapon'

Wait for Weapon halts script execution until a weapon is present a the given (x,y) location.
Via script flags, specific weapons can be detected.

CDbRoom::IsMonsterWeaponTypeAt() has been added to detect specific monster weapons.
UI widgets for Wait for Weapon have been added to CCharacterDialogWidget.
CCharacterDialogWidget::SetBitFlags() has been extended to allow multiple BitFlag lists.

PR #34, opened 2020-09-22
Remove Wait for Weapon script command

Hypexion:
A number of PRs (#6, #9 and #13) were merged to add a new script command, Wait for Weapon. If new script commands require a major version, this work will have to removed for now, and reapplied later.

This PR does the removing part by reverting the PRs that added Wait for Weapon and cleaning up places any other places that refer to it.

Please be aware that #20, which adds Set Behaviour is built assuming these PRs exist, so when they are readded, they must be merged before that work.


mrimer:
Ah, gotcha!  I hadn't realized that previous merges hadn't been released yet, but this makes sense in hindsight. Thank you for being conscientious on this point.  Yes, let's back out these new features for now, and we'll reintroduce them together with Set Behaviour. 

How about I merge this change, and then would you please put up a fresh PR with these feature additions, which we can label to merge into 5.2?

Suggest we work through whichever bugs and non-version-breaking feature changes we'd like to make for 5.1.2, and once we have these released, we'll immediately provide a 5.2 alpha build for testers to begin playing with.

mrimer:
@Hypexion , I'd like to capture other changes that have been made since 5.1.1 was released?

Would you please provide me either a list of commits, or the earliest commit, that you're aware of, which I should review and capture for the 5.1.2 patch?

Hypexion:
> 
> 
> @Hypexion , I'd like to capture other changes that have been made since 5.1.1 was released?
> 
> Would you please provide me either a list of commits or PRs, or the earliest commit/PR, that you're aware of, which I should review and capture for the 5.1.2 patch?
> 
> Edit: Is it just all of the PRs that have been merged in the repo history?

All the changes that were merged are new fixes since the last official release. This should be the only new feature, with the rest being bug fixes.

mrimer:
> All the changes that were merged are new fixes since the last official release. This should be the only new feature, with the rest being bug fixes.

Thank you.  I'll familiarize myself with these fixes.

PR #36, opened 2020-09-22
DROD RPG: Make monster weakness behaviours function on accessories

Hypexion:
Based on this request, I suppose: http://forum.caravelgames.com/viewtopic.php?TopicID=41033

A little refactoring and a change so that adding Goblin Weakness or Wyrm Weakness to custom accessories works. This can be extended to support shields, but there's precedent for shields using behaviours in different ways, so I haven't added them for the moment.

### DIFF
--- a/drodrpg/DROD/GameScreen.cpp
+++ b/drodrpg/DROD/GameScreen.cpp
@@ -1810,6 +1810,8 @@ WSTRING CGameScreen::GetEquipmentPropertiesText(const UINT eCommand)
 					def = (int)pCharacter->getDEF();
 					bMetal = pCharacter->IsMetal();
 					bBeamBlock = pCharacter->HasRayGun();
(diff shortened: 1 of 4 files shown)


mrimer (approved):

mrimer:
@Hypexion , Though you posted the topic as a feature request, the old behavior was probably an oversight and was arguably a bug.

Not sure of the potential impact from making this change in v1.2.8.

Can you provide guidance?  Should we hold off including this change until major release 1.3, or would the community prefer to have this behavior updated in 1.2.8?

Hypexion:
@mrimer If this is considered an oversight/bug, in my view it's a safe change for a minor update. It's not a big change to the gameplay code, and it doesn't change any text resources or hold data.

mrimer:
I appreciate the feedback.  @kieranmillar also weighed in that this change shouldn't affect any published holds, so I'll merge it for 1.2.8.

PR #37, opened 2020-09-23
Fix ExtraVars overriding non-character monster stats in SetMonstersFromExploredRoomData().

mrimer:
As per forum bug topic: http://forum.caravelgames.com/viewtopic.php?TopicID=43560

Also some related code hygiene:

LoadMonster now doesn't populate ExtraVars, as this isn't necessary.
Assert empty ExtraVars for non-character monsters on room import.
Refactor references to monster ExtraVars.
Refactor and remove SetExtraVarsFromMembers.
Rename SetMembersFromExtraVars -> SetMembers.

### DIFF
--- a/drodrpg/DRODLib/DbRooms.cpp
+++ b/drodrpg/DRODLib/DbRooms.cpp
@@ -1403,7 +1403,7 @@ MESSAGE_ID CDbRoom::SetProperty(
 				{
 					BYTE *data;
 					Base64::decode(str,data);
-					pImportMonster->ExtraVars = (const BYTE*)data;
+					const CDbPackedVars vars = (const BYTE*)data;
 					delete[] data;
 					if (pImportMonster->wType == M_CHARACTER)
 					{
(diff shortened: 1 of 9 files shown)


mrimer:
@Hypexion , I'm going to merge this PR to keep things moving fairly quickly, but feel free to review and comment on it, make suggestions, etc.

drodrpg/DRODLib/DbRooms.cpp
```diff
@@ -6235,9 +6238,9 @@ CMonster* CDbRoom::LoadMonster(const c4_RowRef& row)
 	CMonster *pNew = AddNewMonster((MONSTERTYPE)wMonsterType, wX, wY);
 	if (!pNew) throw CException("CDbRoom::LoadMonster: Alloc failed");
 //	pNew->bIsFirstTurn = (p_IsFirstTurn(row) == 1);
-	pNew->ExtraVars = p_ExtraVars(row);
 	pNew->wO = pNew->wPrevO = p_O(row);
-	pNew->SetMembersFromExtraVars();
+	const CDbPackedVars vars = p_ExtraVars(row);
```
mrimer, line 6242: Creates a regression (NPC exports lose data) that is fixed in #260 .

PR #38, opened 2020-09-23
Add VS2019 solution file for DROD RPG 

Hypexion:
There are VS2019 project files for all the DROD RPG components, but currently no VS2019 solution for easy building.

This adds the solution, and also removes the CARAVELBUILD preprocessor flag from the DROD RPG drod project, since Git builds don't have access to the private CaravelNet library.


mrimer:
@Hypexion , Thank you for the PR, and sorry about missing that file.
The solution settings aren't quite right to match the project files, and the dependencies appear to be missing.

I've added and pushed my own proposed PR for this here:
https://github.com/CaravelGames/drod/pull/40

Hypexion:
Closing in favour of #40.

PR #39, opened 2020-09-23
Retain sword state of custom player roles with temporal clones

Hypexion:
Fix for this report: http://forum.caravelgames.com/viewtopic.php?TopicID=44202

Temporal clones determine if their weapon should be active based on their identity. If this is set to a customer character, the identity will never be one that can weild a sword. This is inconsistent with the Swordsman, which uses "appearance" to determine if it can be armed. Appearance is based on the character type the custom character is based on.

By allowing temporal clones to have "appearance", this inconsistency can be fixed, allowing temporal clones of custom roles to use swords.

### DIFF
--- a/DRODLib/TemporalClone.h
+++ b/DRODLib/TemporalClone.h
@@ -58,6 +58,7 @@ public:
 
 	//player stats on creation
 	UINT wIdentity;
+	UINT wAppearance;
 	bool bInvisible;
 	bool bIsTarget;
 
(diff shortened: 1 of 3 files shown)


DRODLib/TemporalClone.h
```diff
@@ -58,6 +58,7 @@ class CTemporalClone : public CMimic
 
 	//player stats on creation
 	UINT wIdentity;
+	UINT wAppearance;
```
mrimer, line 61: Wow!  This change is this simple?  I didn't realize that.  Amazing.

Suggest initializing wAppearance in the constructor method (to a sentinel value) so that it will be apparent if it's not being set properly anywhere during use in the future.

mrimer (approved):

PR #40, opened 2020-09-23
Add VS2019 solution file for DROD RPG

mrimer:


Hypexion:
I can confirm that both DROD RPG and DRODUtil build from this solution file without any changes being required. This looks good to go.

mrimer:
Thank you for taking the time to review and confirm!

PR #44, opened 2020-09-24
DROD RPG: Fix DRODUtil linking

Hypexion:
Even though I look at it every day at my job, appearently I can't read Visual Studio's build output. So I missed that DRODUtil wasn't actually linking for VS2019.

The problems:

- A stub function was missing from GameScreenDummy.h, which has been added
- Some of the release libraries were incorrect
- There is some kind of default library conflict occuring for the DROD RPG version of DRODUtil. For now, I've set it to ignore libcmt, which allows DRODUtil to link.

### DIFF
--- a/drodrpg/DRODUtil/DRODUtil.2019.vcxproj
+++ b/drodrpg/DRODUtil/DRODUtil.2019.vcxproj
@@ -256,7 +256,7 @@
       <Culture>0x0409</Culture>
     </ResourceCompile>
     <Link>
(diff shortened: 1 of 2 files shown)


mrimer:
Thank you for catching these issues.
There are some more build configuration settings to correct, and I'm working on those now.
I might know how to fix the library linker issue by fiddling with configuration settings, which I had to do for the other projects, so I'll address those also.

mrimer:
Closing this to use https://github.com/CaravelGames/drod/pull/47

PR #45, opened 2020-09-24
DROD RPG: Beam Blocker behaviour for equipment

Hypexion:
Sort of fix/feature hybrid, based on the idea from this thread: http://forum.caravelgames.com/viewtopic.php?TopicID=35930

Examination of the code reveals that custom equipment with the Beam Attack behaviour is meant to provide protection from beams. However, due to an oversight, it doesn't not. So I added a new behaviour that allows custom equipment to block beams. Swords will stop beams at their tiles, armor will protect the player if they are facing the beam, and accessories will protect the player. The advantage of using a new behaviour is that it describes what it does, instead of the slightly weird situation where something with "attack" in the name provides a defensive ability.


drodrpg/Texts/MIDs.h
```diff
@@ -1409,6 +1409,7 @@ enum MID_CONSTANT {
   MID_BehaviorAttackInFront = 1526,
   MID_BehaviorAttackAdjacent = 1527,
   MID_BehaviorBeamAttack = 1528,
+  MID_BehaviorBeamBlock = 1748, //New.
```
mrimer, line 1412: When new MIDs are created, we need to include a text hack into CDbBase::GetMessageText(), so that a user who just installs the .exe as a patch without a new main .dat file will be able to see the text.

Same for any other new MIDs that get added.

drodrpg/Texts/MIDs.h
```diff
@@ -1556,7 +1557,7 @@ enum MID_CONSTANT {
   MID_Snow = 1151,
   MID_LightCrossfade = 1278,
 
-  MID_END_UNUSED=1748
+  MID_END_UNUSED
```
mrimer, line 1560: This change doesn't seem correct.  This end MID sentinel should be explicitly set as one above the max MID value in the set.  This should be done automatically when running DRODUtil following these sample steps (I keep these in a .bat file to run whenever I add a new text):

```
#Sample build steps for generating a fresh barebones drod5_0.dat file
cd C:\Caravel\DROD_Dev\DRODUtil\Release
drodutil delete
drodutil create
drodutil import C:\Caravel\DROD_Dev\Texts
```

mrimer (changes requested):
I like it!

There are a couple MID-related bookkeeping issues to address.

drodrpg/Texts/MIDs.h
Hypexion: I've fixed this up manually, and I'll keep an eye out in case this happens again in the future.

mrimer (approved):

mrimer:
Agree about the oversight for custom equipment not blocking beams properly.

For adding a new script command parameter, we should keep that till 1.3 (which, at the current great rate of addressing issues, may come sooner than we think).

mrimer:
I created a [v1.3 branch](https://github.com/CaravelGames/drod/tree/rpg-1.3-dev) where we can start pushing changes for 1.3.
As bug fixes are made to the main branch, we can rebase the 1.3 branch on top of it to keep all the changes in sync.

Would you please edit this PR to go into the 1.3 branch?

PR #58, opened 2020-09-26
Removed bits of code from PNG importing that seemed to reinvent what …

EvidentlyCube:
…SDL can already do (gracefully import PNGs that are palette based and have alpha information for colors) and made it use the SDL's built-in solution that converts those PNGs to ones with alpha channel

----

[Related thread.](http://forum.caravelgames.com/viewtopic.php?TopicID=44751)

@mrimer  Please very carefully decide if that's an acceptable solution. I could only guess as for the reason why that complicated handling for tRNP images was there in the first place (My guess is it either did not work correctly in some older version of SDL, or was an optimization to avoid expanding every single PNG when a faster creation of the surface was possible). I checked the following things with this change:

1. The old images that worked for me still work for me.
2. The old images that did not work for me no work correctly
3. `#C0C0C0` is still the transparent color for character sprites

### DIFF
--- a/FrontEndLib/PNGHandler.cpp
+++ b/FrontEndLib/PNGHandler.cpp
@@ -150,76 +150,31 @@ CS_error:
 	// Grayscale with alpha isn't supported by SDL, so make it RGBA
 	if (colortype == PNG_COLOR_TYPE_GRAY_ALPHA)
 		png_set_gray_to_rgb(pPNG);
-
-	// Got transparency ?
-	if (png_get_valid(pPNG, pInfo, PNG_INFO_tRNS))
-	{
-		png_bytep trans;
-		int num_trans;
-		png_get_tRNS(pPNG, pInfo, &trans, &num_trans, &trans_values);
-
-		// Convert transparency to colorkey and/or surface alpha, if possible
-		if (colortype == PNG_COLOR_TYPE_PALETTE)
-		{
-			int i = num_trans, ck = -1;
-			while (i--) {
-				if (trans[i] == 0)
-					ck = i;
-				else if (surfalpha < 0)
-					surfalpha = trans[i];
-				else if (trans[i] != surfalpha)
-				{
-					surfalpha = -1;
-					break;
-				}
-			}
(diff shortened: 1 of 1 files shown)


mrimer:
Thanks for the clean-up!
IIRC, the PNG implementation was originally provided by Trick, probably back in 2005 or even earlier.
It's quite possible the more streamlined workflow wasn't supported at the time.
Or, maybe, Trick wanted to provide support for color-keyed PNGs.  That's probably not a useful option to us; if we have PNG images in this format, we can just convert them to something more typical anyway.
Let's see how it goes!  We can always revert if there's an issue.

EvidentlyCube:
> Or, maybe, Trick wanted to provide support for color-keyed PNGs.

I don't think PNGs can be really color-keyed from what I see. Again, I am not very well-versed and it's mostly stuff I learned yesterday.

There are three broad categories of PNGs (as per [this page](http://www.libpng.org/pub/png/book/chapter08.html)):
 1. RGB (full 32-bit so each pixel in the image has its own color, without any palette)
 2. Grayscale (so each pixel is at most one byte determining its grayness)
 3. Palette (up to 256 colors)

Here we're only talking about palette-based ones. It seems normally they don't have any transparency information, and the only way to have it is to use tRNS chunk, which can provide alpha channel value for each color in the palette at most, or just a single color at worst. 

The code I removed seems to have been doing the following
1. If there is more than one transparent color then convert image to RGB
2. If there is only a single transparent color or many... do something I don't fully understand (the second bit of the code). Perhaps it is indeed color keying, in which case I don't believe it's ever been used in any hold nor has much value considering you can just use an RGB PNG with full alpha channel.

The bottom line is - I agree with the approach "Let's just remove it and hope it works out correctly" because there is another problem with the existing code. Not only the images draw incorrectly, they also crash the game in certain conditions (I believe rendering them while death fade out is playing).

mrimer:
Thanks for the response.  Apologies, I misspoke; I meant palette-based PNGs.
I agree with everything you're saying here.
Thank you for the fix!

PR #61, opened 2020-09-28
Room preview from clicking on minimap now will have access to the current game's state variables (in a safe way).

EvidentlyCube:
The original code would only set `stats` field of the temporary current game, which is immediattely overwrrited by the left-empty `statsAtRoomStart`. Happened in a call to `CCurrentGame::SetTurn(0)`, followed by `CCurrentGame::SetPlayerToRoomStart()` followed by `this->stats = this->statsAtRoomStart`.

The new solution will instead set `statsAtRoomStart` to the current game's `statsAtRoomStart` effectively exposing the variables in the previews.

### DIFF
--- a/DROD/GameScreen.cpp
+++ b/DROD/GameScreen.cpp
@@ -5989,7 +5989,7 @@ void CGameScreen::ShowRoomTemporarily(const UINT roomID)
 	pTempGame->bSwordsmanOutsideRoom = true;
 	pTempGame->CompletedScripts = this->pCurrentGame->CompletedScripts;
 	pTempGame->ConqueredRooms = this->pCurrentGame->ConqueredRooms;
-	pTempGame->stats = this->pCurrentGame->stats;
+	pTempGame->statsAtRoomStart = this->pCurrentGame->statsAtRoomStart;
 	pRoom->SetCurrentGame(pTempGame);
 	pTempGame->SetTurn(0, CueEvents);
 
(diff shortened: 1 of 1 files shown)


EvidentlyCube:
What I could do instead is set the temporary game's `statsAtRoomStart` to current game's `stats`. The benefit would be that you'd see the room preview in exactly the state you'd see it as if you just went there, which could be meaningful when you did something in a room that you knew triggered a variable changed and would be able to inspect some other room and see how it was affected immediately.

The downside of this solution is that in some extremely rare cases this could be used to get to some knowledge early. Like maybe a room where after causing the variable change you can still get stuck. But is it really a risk? I don't think so.

The more I think about it the more I am convinced it should actually be changed to:

```
pTempGame->statsAtRoomStart = this->pCurrentGame->stats;
```

It avoids awkwardness when you KNOW that something should be changed and everyone accepts it, but the player will see a stale preview until they bother to leave the room. 

In other words, because I feel like I made a word salad, there are two ways to approach this:

1. The one I implemented - room preview acts with variables as if nothing yet happened in the room the player is in right now.
2. The other one - room preview acts as if the player visited the previewed room **right now**.

I think ultimately #2 makes slightly more sense.

EvidentlyCube:
[Relevant Thread](http://forumskell.caravelgames.com/viewtopic.php?TopicID=44758).

mrimer:
> I think ultimately the second option makes slightly more sense.

I agree with your reasoning.  Let's go with the second option.

RPG does a similar thing (or should, anyway), where when you preview another room, the monster stats in that room are based on the current game state, as opposed to showing for the point in time when you entered the current room.

EvidentlyCube:
Updated PR!

mrimer (approved):

PR #62, opened 2020-09-28
Character options dialog will now auto save and close on pressing Ent…

EvidentlyCube:
…er, and close without saving on pressing Escape

[Relevant thread.](http://forumskell.caravelgames.com/viewtopic.php?TopicID=44759&page=0#435468)


mrimer (approved):

mrimer:
I like the hotkeys!

Wondering whether they are/should be documented in the respective help file.

EvidentlyCube:
I think enter/escape hotkeys for save/cancel are ubiquitous enough that it's either obvious or very learnable.

PR #63, opened 2020-09-28
Character processing seq buffer overflow

EvidentlyCube:
I've tested that the changes to `_itoW` work correctly but please double check it. If you're worried about speed I'll gladly create another function that's safe.

[Relevant thread](http://forumskell.caravelgames.com/viewtopic.php?TopicID=44139).

### DIFF
--- a/BackEndLib/Wchar.cpp
+++ b/BackEndLib/Wchar.cpp
@@ -560,7 +560,7 @@ int _Wtoi(const WCHAR* wsz)
 }
 
 //*****************************************************************************
-WCHAR* _itoW(int value, WCHAR* wcz, int radix)
+WCHAR* _itoW(int value, WCHAR* wcz, int radix, int bufferLength)
 {
 	int i = 0;
 	bool bNegative = false;
@@ -568,12 +568,17 @@ WCHAR* _itoW(int value, WCHAR* wcz, int radix)
 	{
 		value = -value;
 		bNegative = true;
+		bufferLength--; // We have one character fewers for numbers because we need negative sign
 	}
+	bufferLength--; // Make space for null termination char
 	do {
 		int val = value % radix;
 		if (val < 10)
 			buffer[i++] = (value % radix) + '0';
 		else buffer[i++] = (value % radix) - 10 + 'A';
+
+		if (i >= bufferLength) // Ran out of space in the buffer
+			break;
 	} while ((value /= radix) > 0);
 	if (bNegative)
 		buffer[i++] = '-';
--- a/BackEndLib/Wchar.cpp
+++ b/BackEndLib/Wchar.cpp
@@ -86,7 +86,7 @@ const WCHAR wszITag[] = { We('<'),We('i'),We('>'), We(0) };
 const WCHAR wszEndITag[] = { We('<'),We('/'),We('i'),We('>'), We(0) };
 const WCHAR wszPTag[] = { We('<'),We('p'),We('>'), We(0) };
 
-static char buffer[512];
+static char buffer[WCHAR_INTERNAL_BUFFER];
 
 //*****************************************************************************
 void AsciiToUnicode(const char *psz, WSTRING &wstr)
@@ -562,6 +562,9 @@ int _Wtoi(const WCHAR* wsz)
 //*****************************************************************************
 WCHAR* _itoW(int value, WCHAR* wcz, int radix, int bufferLength)
 {
+	// Internal buffer's max length should never ever be exceeded
+	bufferLength = min(WCHAR_INTERNAL_BUFFER, bufferLength);
+
 	int i = 0;
 	bool bNegative = false;
 	if (value < 0) //negative?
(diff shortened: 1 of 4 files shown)


BackEndLib/Wchar.cpp
```diff
@@ -560,20 +560,25 @@ int _Wtoi(const WCHAR* wsz)
...
+		bufferLength--; // We have one character fewers for numbers because we need negative sign
 	}
+	bufferLength--; // Make space for null termination char
 	do {
 		int val = value % radix;
 		if (val < 10)
 			buffer[i++] = (value % radix) + '0';
 		else buffer[i++] = (value % radix) - 10 + 'A';
+
+		if (i >= bufferLength) // Ran out of space in the buffer
```
mrimer, line 580: I'm not worried about speed impact from introducing this check.
(I would expect time spent doing the multiple string copies below dominates.)

However, I can't think of a use case where we'd actually want to truncate a numeric conversion because the source buffer isn't large enough.  This is a bug; we should simply resize the input string so the result fits.

Nothing wrong, per se, with keeping the check, though.  Some suggestions:
1. Add an assertion failure if we reach this point.
1a. Swap the if check for an assertion.
2. Change the break to a thrown exception.

Thoughts?

BackEndLib/Wchar.h
```diff
@@ -129,7 +129,7 @@ bool isWInteger(const WCHAR* wcz);
 bool isInteger(const char* pcz);
 
 int      _Wtoi(const WCHAR* wsz);
-WCHAR*   _itoW(int value, WCHAR* wcz, int radix);
+WCHAR*   _itoW(int value, WCHAR* wcz, int radix, int bufferLength = INT_MAX);
```
mrimer, line 132: The buffer used in this function has a magic number size of 512, so bufferLength should never be allowed at a value larger than 512.

DROD/CharacterOptionsWidget.cpp
```diff
@@ -68,23 +68,27 @@ CCharacterOptionsDialog::CCharacterOptionsDialog(
 void CCharacterOptionsDialog::SetCharacter(
 	const CCharacter *pCharacter
 ){
-	WCHAR temp[8];
+	WCHAR temp[10];
```
mrimer, line 71: I'm expecting we need to address this overflow in the rpg codebase as well, correct?

mrimer (changes requested):
Thank you for fixing this!

I think there are a few loose ends we need to tie up (see comments).

DROD/CharacterOptionsWidget.cpp
EvidentlyCube: I don't know if RPG has processing sequence editing - this was introduced late and I do not remember ever porting it.

BackEndLib/Wchar.cpp
EvidentlyCube: The only reason I'd be tempted to truncate rather than crash hard would be to make it slightly more user friendly - a crashing game can make you lose a lot of progress. On the other hand once, it's fixed in a given place once it should never be a problem again. 

I think an assertion beep but gracefully trimming the excess is the most user friendly solution here.

BackEndLib/Wchar.cpp
```diff
@@ -560,20 +560,27 @@ int _Wtoi(const WCHAR* wsz)
...
 	}
+	bufferLength--; // Make space for null termination char
 	do {
 		int val = value % radix;
 		if (val < 10)
 			buffer[i++] = (value % radix) + '0';
 		else buffer[i++] = (value % radix) - 10 + 'A';
+
+		if (i >= bufferLength) {
+			ASSERTP(i < bufferLength, "Converting a string to a number ran out of space in the provided buffer.");
```
mrimer, line 581: We don't need the condition check here, because it's always false.
Suggest:
ASSERT(!"_itoW: Converting a string to a number ran out of space in the provided buffer.");

BackEndLib/Wchar.h
```diff
@@ -129,7 +129,7 @@ bool isWInteger(const WCHAR* wcz);
 bool isInteger(const char* pcz);
 
 int      _Wtoi(const WCHAR* wsz);
-WCHAR*   _itoW(int value, WCHAR* wcz, int radix);
+WCHAR*   _itoW(int value, WCHAR* wcz, int radix, int bufferLength = 512);
```
mrimer, line 132: This is closer.  One more loose end:
In the method definition, we need to clamp bufferLength to 512.  Suggest replacing these magic number instances of 512 with a #define or something even better for ease of maintenance.

mrimer (changes requested):

DROD/CharacterOptionsWidget.cpp
mrimer: Ah, I didn't look closely enough here.  Didn't realize this was specific in the character options dialog box.
I agree this isn't relevant to RPG.

BackEndLib/Wchar.cpp
EvidentlyCube: I completely forgot you can use `ASSERT()` like that, been a while since I touched this codebase :). Posted an update

mrimer (approved):

BackEndLib/Wchar.cpp
mrimer: No worries!  Same here :-)

PR #64, opened 2020-09-28
T-Layer covered items (fuse/token under Mirror/keg) will now be descr…

EvidentlyCube:
…ibed when right-clicking

[Relevant thread.](http://forumskell.caravelgames.com/viewtopic.php?TopicID=43670)

### DIFF
--- a/DROD/RoomWidget.cpp
+++ b/DROD/RoomWidget.cpp
@@ -1532,6 +1532,19 @@ void CRoomWidget::DisplayRoomCoordSubtitle(const UINT wX, const UINT wY)
 		}
 	}
 
+	// T-Covered tiles.
+	const UINT tCoveredTile = this->pRoom->GetCoveredTSquare(wX, wY);
+	if (tCoveredTile != T_EMPTY)
+	{
+		switch (tCoveredTile)
+		{
+			case T_TOKEN:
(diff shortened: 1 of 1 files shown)


DROD/RoomWidget.cpp
```diff
@@ -1532,6 +1532,19 @@ void CRoomWidget::DisplayRoomCoordSubtitle(const UINT wX, const UINT wY)
...
 
+	// T-Covered tiles.
+	const UINT tCoveredTile = this->pRoom->GetCoveredTSquare(wX, wY);
+	if (tCoveredTile != T_EMPTY)
+	{
+		switch (tCoveredTile)
+		{
+		case T_TOKEN:
+			mid = GetTokenMID(this->pRoom->GetTParam(wX, wY)); break;
+			default: mid = TILE_MID[tCoveredTile]; break;
```
mrimer, line 1543: Indentation here is tricky to read, but code is functionally correct.

mrimer (approved):

DROD/RoomWidget.cpp
EvidentlyCube: You're right. I've updated this to be consistend with the switch case formatting used throughout the function

mrimer (approved):

PR #65, opened 2020-09-28
While loop scripting commands

Hypexion:
Related thread: http://forum.caravelgames.com/viewtopic.php?TopicID=40234

This PR adds `While ...` and `While End` commands to DRODscript. This are paired together to create label-less script loops. The implementation primarily uses the logic from `If ...`, with some additions to account for the commands having slightly different needs. Pretty printing is supported! These commands have the same colours as the `If ...` family of commands, but other colours are possible.

At the moment, there's no `Break` or `Continue` command, but if others think they would be valuable, I'll add them.


Data/Help/1/script.html
```diff
@@ -401,6 +401,8 @@ <h3><u>Character Script Commands</u></h3>
         <li>See "<a href="#branchingexamples">Examples of branching code
              structures</a>" below.</li>
       </ol></li>
+  <li><a name="while"><b>While ...</b></a> - This command functions like <a href="#if"><b>If ...</b></a>, and is paired with a <a href="#whileend"><b>While End</b></a> to create a loop.</li>
```
EvidentlyCube, line 404: Can you expand the explanation for those who are not programmers? I think maybe something like this would be a bit more beginner friendly:

> "While" and "While End" function exactly like "If/If End" with one exception - when "While End" is reached, the script returns back to its matching "While" command, effectively rerunning the inner script block repeatedly until the condition is false.

Or just take pieces of it.

EvidentlyCube (changes requested):
I'd like for this to have a bunch of unit tests but let's defer that until I fix the unit tests

EvidentlyCube:
You're a blessing among us. Now I need "Repeat for every tile at rectangle..."

Data/Help/1/script.html
Hypexion: Yeah, this is a better description. I'll have a bit of a think about it, but I'll probably go with this.

mrimer:
This is a great feature!  Love the idea of looping without labels.
I'm not clear on (at least) a couple types of interactions, or at least I need to ask about them:
1. how gotos and while loops interact.  That is, while executing inside a while loop, does a goto exiting the loop work properly?
2. how if blocks and while loops interact if they are (a) nested, and (b) partially nested/overlapping, i.e., If -> While -> If End -> While End.  Is (a) supported?  Is (b) an error?

Hypexion:
> 
>  how gotos and while loops interact.  That is, while executing inside a while loop, does a goto exiting the scope of the loop work as expected?  I guess I mean I would expect the looping is cancelled?  (Or, rather, am I correct in assuming that on script execution reaching a While End command, the code simply scans backwards/up the script commands till finding the most recent While command, and jumps there?)

Your assumption is lamost correct - when the script execution hits a `While End` it scans backwards to fine a matching `While` command. This is implemented in a similar way to how `If` blocks are jumped over, but backwards. DRODscript doesn't have any concept of being "inside a loop" so jumping out with a `Goto` will simply move the command index somewhere else, without ill effect.

> 
> Are nested whiles supported?

Assuming I haven't messed something up (which is always possible), nesting loops should work properly, much like how `If` blocks can be nested.

> 
> how if blocks and while loops interact if they are (a) nested, and (b) partially nested/overlapping, i.e., If -> While -> If End -> While End.  Is (a) supported?  Is (b) an error?

If everything is done correctly, regular mixed nesting should work. However, now that you've brought it up, it is possible to "splice" `If` and `While` blocks in the way you've mentioned. Currently, the scanning for `If` and `While` ignore each other's commands, so it will work, just not in a way that's going to be nice. Also it'll mess up the pretty printing. I'll have a look to see if this situation can at least get an error marking.

EvidentlyCube:
One interesting thing I just realized is that you don't need `while` at all - you could use regular `if` and have a different ender, say `If End-Loop`. This would of course bring its own set of problems: `If -> Else If -> If End-Loop`.

Or maybe a different way - Keep `if` and `if end`  but have a a command `Goto last If`, which would just scan back to the last If or ElseIf. From the perspective of the casual user `Goto last if` could be the easiest to understand, thoughts?

Also, @Hypexion, PR that fixes tests is up #69 , once it's merged I'd love if you could add tests for the While command (or whatever it ends up being).

mrimer:
> Or maybe a different way - Keep `if` and `if end` but have a a command `Goto last If`, which would just scan back to the last If or ElseIf. From the perspective of the casual user `Goto last if` could be the easiest to understand, thoughts?

I like your idea of maintaining a single looping construct in order to avoid If/while loop splicing.

As I think about this, it seems like this flow would always necessarily jump back to If - but not IfElse - in order to have the expression reevaluated, to decide whether to reenter the then or else block next iteration.

Do we lose anything by defining labelless expression-dependent iteration in this fashion?

EvidentlyCube:
Another idea!

A command `Goto smart...` which has these options to select from:

 - `Last If or Else If` - If inside an `Elseif` block it would just jump back to the `Elseif` section, if inside `If` block would jump back to the `if`
 - `Last If`  - Regardless of being in side `Elseif` or `If` block, it goes back to the `If` at the beginning of the `if-elseif` tree.
 - `Last If or Else If (conditionless)` same as `If or Else If` but ignores the condition check and runs the block again
 - `Last If` ditto
 - `Next elseif/else` - finds the next ElseIf/Else in the current block and runs it (checking the condition in case of `ElseIf`)
 - `Next elseif/else (conditionless)` the same except assumes the condition is true for `Elseif` - kinda like fallthrough switch-case

mrimer:
Thanks for the suggestions!
Those options are a bit confusing to me.
There is already a conditionless if/elseif, in the form of a label.

EvidentlyCube:
> There is already a conditionless if/elseif, in the form of a label.

Sure, but the same argument could be said about implementing While - you can already do that with labels. I am not necessarily saying the ideas I provided should be implemented (I do like the idea of falling through to the next elseif block, just like falling through to the next switch case), but if instead of adding a new "Do this one specific thing" command we do "Generic control flow command" that has a list of options, we make things a bit more scalable. Next time we want to implement something similar we don't need to pollute the already sizeable list of commands.

mrimer:
I think I see your point.

The difference with while or a "jump back to if" command is that it's simpler (fewer instructions) than the current workflow.

These other commands don't simplify the workflow beyond what's already supported.

Hypexion:
I definitely see the appeal of an advanced `Goto` command, especially once that's not bouncing around to different labels. Given the issue of splicing, a command that can "bounce back" to a previous `If` (or `Else If`) command provides a cleaner solution. I can put together a proof-of-concept for this based on the `While End` implementation that goes back to an `If` pretty easily.

The conditionless versions seem less useful on first glance, but if you can provide an example of why you'd want to use them, I'll be happy to add those options in as well.

EvidentlyCube:
> The conditionless versions seem less useful on first glance, but if you can provide an example of why you'd want to use them, I'll be happy to add those options in as well.
I probably encountered a case where each of those commands would've been useful at least once, but not being able to recall specifics must mean they were not a big hassle to write differently. So probably not worth the time implementing them, except...

"Fall through to the body of the next If/Else". Quite often I find myself having to do something like this:
```
If ...
      Wait for item Floor
   <do something>
Elseif ...
      Wait for item Firetrap
   <do the same thing>
Elseif ...
      Wait for item <whatever>
   <once more>
Elseif ...
  <repeated 10 more times>
Ifend
```

And when I need to change what happens in `<do something>` I have to make the same change everywhere. With "Fall through" command I could have the logic in one place, without having to use separate labels, go subs and returns.

mrimer:
@EvidentlyCube , I just noticed I edited your comment instead of quoting it, like I intended.  Whoops!  My sincere apologies.  I hadn't intended to modify your comment.

Hypexion:
Closing in favour of #133

PR #74, opened 2020-09-29
Walking down stairs will no longer happen on turn 0

EvidentlyCube:
[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=44764)

### DIFF
--- a/DRODLib/CurrentGame.cpp
+++ b/DRODLib/CurrentGame.cpp
@@ -6235,6 +6235,9 @@ void CCurrentGame::ProcessPlayer_HandleLeaveLevel(
 	if (CueEvents.HasAnyOccurred(IDCOUNT(CIDA_PlayerDied), CIDA_PlayerDied))
 		return;
 
+	if (this->wTurnNo == 0)
+		return;
+
 	//Write a demo record if recording.
 	if (this->bIsDemoRecording)
 	{
--- a/DRODLib/CurrentGame.cpp
+++ b/DRODLib/CurrentGame.cpp
@@ -6235,7 +6235,7 @@ void CCurrentGame::ProcessPlayer_HandleLeaveLevel(
 	if (CueEvents.HasAnyOccurred(IDCOUNT(CIDA_PlayerDied), CIDA_PlayerDied))
 		return;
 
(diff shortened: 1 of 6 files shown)


DRODLib/CurrentGame.cpp
```diff
@@ -6235,6 +6235,9 @@ void CCurrentGame::ProcessPlayer_HandleLeaveLevel(
 	if (CueEvents.HasAnyOccurred(IDCOUNT(CIDA_PlayerDied), CIDA_PlayerDied))
 		return;
 
+	if (this->wTurnNo == 0)
```
mrimer, line 6238: Note this change completely skips the action of a script command that goes to a level entrance on turn 0.

If we want to prevent scripts from going to a new level entrance on turn 0, I propose we have the script pause and wait till the next turn to execute this command, as with the other turn 0 actions we've limited via scripting.

Suggest `ProcessPlayer_HandleLeaveLevel` return false in this case, and plumb that return value back up the call stack to where it is called in CCharacter::Process.  If `GotoLevelEntrance` returns false, call `STOP_COMMAND;`

mrimer (changes requested):

DRODLib/CurrentGame.cpp
EvidentlyCube: Will check it!

EvidentlyCube: I've checked and Characters already had a check for that, so I decided to leave `ProcxessPlayer_HandleLeaveLevel` as is. 

I changed one thing however - previously Character was using `wTurnNo` instead of `wPlayerTurn` to determine if it's 0 turn or not, for the purpose of Go to level entrance. I changed it to the latter, because this way it won't auto trigger after double placement. It will wait for the real turn 1. I don't believe it should ever be an issue in any hold, but just in case I'll post about it in the forums so that people may prove me wrong, it'll be trivial to roll it back.

mrimer (approved):

DRODLib/CurrentGame.cpp
mrimer: Thanks for checking and confirming scripted behavior, and for the fix and tests!

PR #90, opened 2020-10-01
Set player weapon myscript override

EvidentlyCube:
…eapon

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=42307)


EvidentlyCube:
See #93 instead

PR #100, opened 2020-10-01
Fix various issues to allow compilation on OSX

TomBrouws:


BackEndLib/Browser.cpp
```diff
@@ -167,6 +167,7 @@ static bool OpenSystemBrowser(
 
 #elif defined(__APPLE__)
 #include <string>
+#include <cstring>
```
EvidentlyCube, line 170: Why this change?

Master/Darwin/CompilingDrod_OSX_2020.md
```diff
@@ -0,0 +1,23 @@
...
+* Install g++ version 4.9 through Homebrew, and point CC and CXX in the Makefile to that version
+* Get gcc v3 pod_char_traits.h from the internet
+* Remove -DCARAVELBUILD compiler flag due to missing CaravelNet source
+* Instead of relying on the `install_name_tool` commands in the Makefile I used macdylibbundler to bundle dylibs correctly (https://github.com/auriamg/macdylibbundler) 
+
+Library versions used:
+* Metakit 2.4.9.7
+* SDL2 2.0.12
+* SDL2_mixer 2.0.4
+* SDL2_TTF 2.0.13 (2.0.15 did not work for me, as the included FreeType version needs a version of libpng that is incompatible with the version used in DROD)
```
EvidentlyCube, line 14: Maybe we should update libpng?

BackEndLib/Browser.cpp
TomBrouws: Because `strlen` used in this OSX specific function is defined in `cstring` so we need to include this in addition to `string`.

Master/Darwin/CompilingDrod_OSX_2020.md
TomBrouws: Using a newer version of libpng resulted in all sorts of new issues and solving those I think is outside the scope of producing a working OSX build

mrimer (approved):

PR #102, opened 2020-10-01
fix: don't draw wading tiles for Young Halph

TomBrouws:
This PR fixes http://forum.caravelgames.com/viewtopic.php?TopicID=39612.

As Halph and Young Halph share graphics they were both marked as having wading graphics. However, Young Halph cannot wade, but wading graphics were still being drawn when Young Halph is killed by being pushed into shallow water.


EvidentlyCube (approved):

mrimer (approved):

mrimer:
Thank you, Tom!

PR #108, opened 2020-10-02
Rebase tss-5.2-dev

mrimer:


mrimer:
@Hypexion @EvidentlyCube , I performed a rebase of 5.2 onto master.  For some reason, however, I got some of the more recent commits but not all of them.  Don't know how I missed them during my process, but I can try again to include them maybe tomorrow.

Tangentially, it's becoming a pain to iteratively rebase 5.2 onto master, because there are merge conflicts to resolve in most commits that I have to re-resolve each time I go through the process.  Can you think of a cleaner way to manage keeping 5.2 on the tip of the main branch?

Hypexion:
@mrimer It'll probably be better to merge in the latest version of master, yes. While rebasing does have it's uses, if it's going to cause lots of conflicts it's probably best not to use it in this case.

Hypexion:
@mrimer Okay, I've had a closer look at this PR, and now I'm certain we want to be merging. Whatever you're doing here is consistently introducing breaking changes into the 5.2 branch. In fact, if nothing has been merged into the 5.2 branch since this PR, my advice would be to revert the whole thing and see if a merge works better.

mrimer:
@Hypexion , okay, I'll try combining newer changes this other way.  I'll try to rewind this branch to our last substantive commits and take it forward again, merging in PRs to master, from that point.

mrimer:
For our records, I reset the 5.2 branch to this commit
https://github.com/CaravelGames/drod/commit/53e6943787d3c83f46f871c82c1a1055f3e2e2bd
Now I'll merge new commits in master onto this branch, resolve and conflicts, and push.

mrimer:
@Hypexion , one last comment here.
I believe I've successfully cleaned up the junk I've introduced to the 5.2 branch and merged the current master onto it.
I think all is in order to resume merging next-version PRs onto 5.2.

I also merged master into rpg-1.3-dev.  For the rebase I did earlier, it has some (not too many) superfluous commits in the middle of the commit history that would require some splice-fu to excise completely.  I don't think it's a big deal to leave the 1.3 branch history as-is and move forward from this point (please correct me if you think I'm wrong about this).

Hypexion:
@mrimer Since the RPG 1.3 branch builds, it's probably fine to leave them. I'm not a git expert, but I think messing around with splice isn't worth the risk.

PR #115, opened 2020-10-02
Fix missing () on function call in Mimic.cpp

Hypexion:
Fix for a build problem. This also occurs in the 5.2 branch, so an additional merge will be required


mrimer (approved):

mrimer:
Thank you! Merged with 5.2.

PR #117, opened 2020-10-02
Modified temporal projection roles. Wubba temporal projection is now …

EvidentlyCube:
…resistant to body attacks + Power token now gives abilities to temporal projections that are already active (there was a bug with stalwart role, which would inherit shallow water traversal from player IF player touched power token before recording, but would not do so if the token was activated during playback)

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=40546)

### DIFF
--- a/DRODLib/DbRooms.cpp
+++ b/DRODLib/DbRooms.cpp
@@ -4088,6 +4088,7 @@ void CDbRoom::ActivateToken(
 					{
 						CTemporalClone *pClone = DYN_CAST(CTemporalClone*, CMonster*, pMonster);
 						pClone->bIsTarget = true;
+						pClone->SetMovementType();
 					}
 					break;
 					default: break;
--- a/DRODLib/Slayer.cpp
+++ b/DRODLib/Slayer.cpp
@@ -28,6 +28,7 @@
 //Implementation of CSlayer.
 
 #include "Slayer.h"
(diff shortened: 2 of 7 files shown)


DRODLib/Slayer.cpp
```diff
@@ -111,6 +112,7 @@ void CSlayer::Process(
 	if (!GetTarget(this->wTX, this->wTY, false))
 		return;
 
+	const CMonster* pTarget = this->pCurrentGame->pRoom->GetMonsterAtSquare(this->wTX, this->wTY);
```
mrimer, line 115: This line appears unnecessary.  Remove.

mrimer (approved):
Thank you!
I noted one line to remove.

DRODLib/DbRooms.cpp
```diff
@@ -4088,6 +4088,8 @@ void CDbRoom::ActivateToken(
 					{
 						CTemporalClone *pClone = DYN_CAST(CTemporalClone*, CMonster*, pMonster);
 						pClone->bIsTarget = true;
+						pClone->SetMovementType();
+
```
mrimer, line 4092: Remove blank line.

EvidentlyCube:
@mrimer Pushed fixes!

mrimer (approved):

PR #127, opened 2020-10-03
fix: set default black color in different manner to ensure it is always set correctly

TomBrouws:
The way the default black color was set in some instances caused issues on OS X, where the color would not always be set correctly, resulting in text not being visible in list boxes. I've changed the constructors using this logic to fix the issue.

### DIFF
--- a/DROD/Chat.h
+++ b/DROD/Chat.h
@@ -67,9 +67,8 @@ namespace CNetChat
 
 	struct Data
 	{
-		Data() : chatID(0), opType(0), timestamp(0), bAdminMessage(false), bAFK(false)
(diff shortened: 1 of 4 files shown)


EvidentlyCube:
Interesting! Do you know why this version works but the old one didn't? Was it because the old code did not initialize the fourth property of color?

TomBrouws:
Interesting indeed. It definitely has something to do with the `a` value not being set, but that doesn't explain why it ^did^ work for some list elements but not for others, in a random fashion. My guess is that because `a` is undefined, the random garbage memory it points at is sometimes interpreted as fully opaque and sometimes as fully transparent by the SDL library.

mrimer (approved):
Thank you!
Yeah, I'm not surprised by odd issues like this on OS X.  Appreciate the fix!

PR #128, opened 2020-10-03
More music variety in the editor

kieranmillar:
Adds a new ini line for each style to customize tracks played in the editor, by default this is the Puzzle then Attack tracks.

No need for this to go into RPG as it already does the equivalent of this.

There are two consequences to this change:
- Any non-official styles the player currently has will play Deep Spaces tracks in the editor unless the user manually adds in a line in their ini files.
- CaravelNet automatic importing of styles when downloading a hold will need to be updated to add in editor lines with suitable defaults when it downloads them.


mrimer (approved):

mrimer:
Neat! Thanks!

PR #132, opened 2020-10-03
Armed monsters that are drawn with transparency automatically mark th…

EvidentlyCube:
…eir tiles as dirty to prevent multiple delta-draws from stacking and causing flickering + ditto for their swords + damaged sword tiles are no longer redrawn if they belong to a hiding clone

I hope you can make sense of it Mike.

[Relevant thread](http://forum.caravelgames.com/message.php?Action=NewReplyMessage&BoardID=7&TopicID=38214)


mrimer (approved):

mrimer:
Oh, neat!  Thank you so much for addressing this annoying flickering!

PR #133, opened 2020-10-04
"Goto Smart" script command

Hypexion:
`Goto Smart` allows labelless jumps to specific script locations.  This can be used to create more compact loops, switch statement like "drop-throughs" and other flow control operations. Inspired by conversations on #65 

This is currently a draft PR to get some feedback and hear thoughts before committing to stuff like writing tests.


DRODLib/CharacterCommand.h
```diff
@@ -184,6 +184,17 @@ namespace ScriptFlag
 		Stalwart=4
 	};
 
+	enum GotoSmartType
```
mrimer, line 187: @EvidentlyCube , I realize I'm not sure whether you were being facetious about actually adding all of these potential ways to jump around in the code.  Do you think it's a good idea to actually have all of these?

mrimer (commented):
The code all looks good to me.  Thanks for prototyping it!

To ensure I understand, if we were to go ahead with this approach, then would we abandon the while loop construct, or would you want to add it as well?

I'd like to double check that @EvidentlyCube , you and I are all thinking it makes sense to have all these options.

I guess I don't have strong feelings either way, though some of them are completely unfamiliar to me and I can't think of use cases for them off the top of my head.  They might be confusing to use in practice, but of course players don't have to use them if they don't understand them.

DRODLib/CharacterCommand.h
mrimer: @Hypexion , my kneejerk reaction here would be to add these as special/predefined options to the "Goto <label>" command.  That is, rather than having a new command, add these as options in the Goto listbox (probably with negative values instead of positive).

Reason being it reads much easier if the command says "Goto LastIf" instead of "Goto Smart LastIf".  Also, we don't need another command, really.

I'm not trying to be prescriptive here; just wanting to start a conversation to see what you've already considered as you designed and implemented this functionality.

DRODLib/CharacterCommand.h
```diff
@@ -184,6 +184,17 @@ namespace ScriptFlag
...
+	{
+		LastIf = 0,
```
mrimer, line 189: I'd refer to "Last*" as "Previous*" to avoid ambiguity in whether this means jumping to the last (as in "final") If in the script.

Hypexion: Good catch. This definitely a thing where I know what context "Last" is being used in that might be not be clear to a user.

DRODLib/CharacterCommand.h
Hypexion: I think there's going to be a tradeoff between keeping the number of commands as low as possible, and the engineering issues that might come up with using the existing Goto structure. The command execution side probably isn't too bad, but I think the UI side will run into issues. Currently, the widget for selecting Goto labels is also used when selecting labels for question options, and I'm not sure if we want the functionallity offered through that path.

EvidentlyCube:
**An opinion that should be popular:** We should make the command selection dialog MUCH taller, like make it as tall or almost as tall as the commands list. UX will skyrocket.

The ones I find very useful off the bat are:

1. `Last If` Is the While equivalent
2. `NextElseOrElseIf (skip conditions)` allows fall-through switch cases (see SCRIPT A. below), which makes it possible to reduces the repetition in the script.

As for the rest - I thought about them since I introduced the idea the first time and I have to admit I didn't come up with any very useful cases that wouldn't be better off using labels. It's probably the best idea to keep the two above, unless others have some suggestions. I do like the format of using a command with selectable options rather than separate commands for each, because it makes the solution more scalable. I am ambivalent about using the good old Goto for this. It'll be a pain from the POV of coding the UI, but from UX perspective I think it'd be fine.

```
SCRIPT A:
     If ... 
           Wait for item Briar growth,0,0,0,0
        Attack tile 0,0,Explosion
     Else If 
           Wait for item Briar root,17,23,17,23
        Attack tile 0,0,Explosion
     Else If 
           Wait for item Withered briar,17,24,17,24
        Attack tile 0,0,Explosion
```

DRODLib/CharacterCommand.h
mrimer: > ...I think the UI side will run into issues. Currently, the widget for selecting Goto labels is also used when selecting labels for question options, and I'm not sure if we want the functionality offered through that path.

Ah, I'd forgotten about that.  I agree it probably wouldn't make sense to populate question options with the smart goto options.

Hypexion:
I've done a rework this functionallity, folding it into the main `Goto` command. I've also pruned down the number of options. There's still some UI work that needs doing, but you can now build this and get the smart goto options for `Goto` and `GoSub`. (I haven't tried with answers and will probably not allow them to used that way.)

The side effect here is that order for labels to negative values, they need to be treated as signed integers instead of unsigned integers. This technically cuts the number of available labels in half, although I doubt any script is using more that 2^31 labels.

DRODLib/Character.cpp
```diff
@@ -5324,17 +5324,113 @@ void CCharacter::Disappear()
...
+// NO_LABEL if not in a block
+int CCharacter::GetIndexOfPreviousIf(const bool bIgnoreElseIf) const
+{
+	UINT wCommandIndex = this->wCurrentCommandIndex;
+	UINT wNestingDepth = 0;
+	bool bScanning = true;
+
+	do {
+		--wCommandIndex;
+		if (wCommandIndex == -1) //No matching if found
```
mrimer, line 5370: I'm guessing there's a compiler warning here.

mrimer (approved):

mrimer:
@Hypexion , I like the newest version. Thank you!  I'm good with merging as-is.

@EvidentlyCube , how do you feel about this approach?

EvidentlyCube:
From POV of usability I'd suggest these changess:
1. Enclose those special labels in `<>` brackets to clearly indicate they are special
2. Give them different color (red?)
3. Put them either at the beginning or at the end of the list - this will happen automatically after doing #170 
4. In the script itself strip the quotation marks for the special labels. 

From POV of code I'll review it now!

DROD/CharacterDialogWidget.cpp
```diff
@@ -4261,6 +4258,28 @@ const
 	return wstr;
 }
 
+WSTRING CCharacterDialogWidget::GetGotoDestinationName(
```
EvidentlyCube, line 4261: If you decide to follow my [suggestion](https://github.com/CaravelGames/drod/pull/133#issuecomment-708294018) it'll likely be best to pass `WSTRING` directly to this function and rename it to `AppendDestrinationName` or something.

DROD/CharacterDialogWidget.cpp
```diff
@@ -5067,7 +5089,7 @@ void CCharacterDialogWidget::SetActionWidgetStates()
 {
 	//Code is structured in this way to facilitate quick addition of
 	//additional action parameters.
-	static const UINT NUM_WIDGETS = 52;
+	static const UINT NUM_WIDGETS = 53;
```
EvidentlyCube, line 5092: Is this increase required now?

DRODLib/Character.cpp
```diff
@@ -5324,17 +5324,113 @@ void CCharacter::Disappear()
 }
 
 //*****************************************************************************
-int CCharacter::GetIndexOfCommandWithLabel(const UINT label) const
+int CCharacter::GetIndexOfCommandWithLabel(const int label) const
```
EvidentlyCube, line 5327: Might be better to pass the index of current label rather than loading it from current state, but I am not 100% sure.

DRODLib/Character.cpp
```diff
@@ -5324,17 +5324,113 @@ void CCharacter::Disappear()
...
 			const CCharacterCommand& command = this->commands[wIndex];
 			if (command.command == CCharacterCommand::CC_Label &&
-					label == command.x)
+				label == command.x)
 				return wIndex;
 		}
+	} else if (label < 0) {
+		const ScriptFlag::GotoSmartType eGotoType = (ScriptFlag::GotoSmartType) label;
+		switch (eGotoType) {
+		case ScriptFlag::GotoSmartType::PreviousIf:
```
EvidentlyCube, line 5341: Rest of the code has `case` indented by one tab.

DRODLib/Character.cpp
```diff
@@ -5324,17 +5324,113 @@ void CCharacter::Disappear()
...
+	bool bScanning = true;
+
+	do {
+		--wCommandIndex;
+		if (wCommandIndex == -1) //No matching if found
+			break;
+
+		CCharacterCommand command = this->commands[wCommandIndex];
+		switch (command.command) {
+		case CCharacterCommand::CC_If:
```
EvidentlyCube, line 5375: Ditto with indent.
Sidenote: why not just return `wCommandIndex`, remove `bScanning` completely and move the above `if (...) break` to the while's condition?

DRODLib/Character.cpp
```diff
@@ -5324,17 +5324,113 @@ void CCharacter::Disappear()
...
+	bool bScanning = true;
+
+	do {
+		++wCommandIndex;
+		if (wCommandIndex == this->commands.size()) //No matching else found
+			break;
+
+		CCharacterCommand command = this->commands[wCommandIndex];
+		switch (command.command) {
+		case CCharacterCommand::CC_If:
```
EvidentlyCube, line 5412: Switch-case intend

EvidentlyCube (approved):

mrimer:
> From POV of usability I'd suggest these changess:

I like all these changes!

Hypexion:
I've tidied things up, and added angular brackets to the special gotos. I decided not to change the colours, since I think it looks pretty obvious that these commands are special now.

EvidentlyCube:
Looking good! Only thing left to do is tests.

mrimer (commented):
Completed my review.  It looks nice!  Agree we should have tests.  This is going to be a nice, solid feature that makes scripting this much easier to deal with.  Thank you!

mrimer (approved):

mrimer:
@EvidentlyCube , if you're good with merging this PR, how about we merge the rebase and then commit this on top of it, or would you prefer the other way around?

mrimer:
Any objections to me merging this PR for 5.2?

EvidentlyCube (approved):

EvidentlyCube:
I am good!

PR #136, opened 2020-10-05
fix: check if monster in square where stun effect is being applied is actually stunned

TomBrouws:
Turns out the methods that add stun effects didn't actually check if the monster at that location is stunned, it just checks if it's alive. If a character is pushed to a square and then immediately killed, `pMonster` is actually the killing monster and a stun effect should not be drawn. The explicit `isStunned` check remedies this.


EvidentlyCube (approved):

mrimer (approved):

mrimer:
[Forum topic](http://forum.caravelgames.com/viewtopic.php?TopicID=38762)

PR #137, opened 2020-10-05
Moved when ResetForPaint() happens so that creating effects for CueEv…

EvidentlyCube:
…ents can still inspect room's previous state

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=39404)

@mrimer  I am not 100% confident this change won't break anything but I think it makes sense to do it like that.


mrimer:
I *think* your change won't break anything, but I'm not 100% confident either.

Let's try it.

I couldn't get the display issue that was being mentioned in the forum topic to manifest from the linked sample hold, though.  Maybe I missed something...

mrimer:
@EvidentlyCube , I'm not sure how to describe this change in a bug fix summary on the forum.  Suggestions?

PR #149, opened 2020-10-08
A big rewrite of FaceWidget.

EvidentlyCube:
Why: I was chasing a bug, where the game would not switch back from custom image avatar to a built-in avatar if you changed player roles. Then when I fixed it something else broke. And then I fix what broke and yet another thing broke. Then I realized the way FaceWidget is written and used is flawed in two ways.

Flaw 1 - trying to fit as much state in as few variables. There was the whole thing with locking moods which was made to not have to have separate set of variables for speakers, which was the main source of issues.

Flaw 2 - a lot of logic that belonged in `FaceWidget` was spread around `GameScreen` , specifically figuring out how to manipulate the moods and the locks.

Analysis - when you think about it, there are essentially three differnt things displayed in the widget. Player's face, Speaker's face and Dying face. Technically Dying face is Player's face but because it behaves slightly differently (it has no special state beyond 'I am active') I think it makes sense to keep it separate to avoid issue.

Solution - there is a new struct that holds Face's state, and all the logic in `GameScreen` that previously had to do weird stuff to figure out how to manipulate the state to get the desired effect just directly tells the widget what to do and which face layer to modify. Speech on start enables Speech face, and disables it when speech ends. Ditto for dying. I think the solution is much more elegant, though there are likely ways to improve it still.

----

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=38410)

### DIFF
--- a/DROD/FaceWidget.cpp
+++ b/DROD/FaceWidget.cpp
@@ -131,18 +131,16 @@ CFaceWidget::CFaceWidget(
 	const int nSetX, const int nSetY,               //    constructor.
 	const UINT wSetW, const UINT wSetH)             //
 	: CWidget((WIDGETTYPE)WT_Face, dwSetTagNo, nSetX, nSetY, wSetW, wSetH)
-	, eSpeaker(Speaker_Beethro)
-	, eMood(Mood_Normal), ePrevMood(Mood_Normal)
-	, dwDelayMood(0), dwStartDelayMood(0)
-	, bMoodDrawn(false), bIsReading(false), bIsBlinking(false), bIsSleeping(false)
-	, bDoBlink(false), bIsDeathAnimation(false)
-	, eMoodSEID(SEID_NONE)
-	, bMoodLocked(false)
 	, nPupilX(0), nPupilY(0), nPupilTargetX(0), nPupilTargetY(0)
 	, dwLastFrame(0), bAlwaysPaintFull(false)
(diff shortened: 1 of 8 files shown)


EvidentlyCube:
And here is a hold I used to test things.

[Speech Tests.zip](https://github.com/CaravelGames/drod/files/5350822/Speech.Tests.zip)


Note - there is one change in behavior. Previously speaking character would be prioritized over dying animation, but I've accidentally made it show the dying animation instead. I don't mind either way so I can either keep it or change it.

DROD/FaceWidget.cpp
```diff
@@ -154,157 +152,199 @@ CFaceWidget::~CFaceWidget()
...
+		default:
+			return &facePlayer;
+	}
+}
+
+const FaceWidgetLayer CFaceWidget::GetActiveLayer() const{
+	if (faceSpeaker.bIsActive)
+		return faceSpeaker.eLayer;
+
+	else if (faceDying.bIsActive)
```
mrimer, line 192: Can remove the two elses here to make the code a bit cleaner and easier to read.

mrimer (approved):
I concur the face logic was difficult to maintain in its previous form.
Thank you for improving it!
I agree your refactoring is a more elegant way to manage the logic and data.  Much appreciated!

EvidentlyCube:
Updated the commit

mrimer (approved):

PR #151, opened 2020-10-08
Pushed slayer who are also seeking will extend wisp from their push p…

EvidentlyCube:
…osition

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=41350)


mrimer (approved):

mrimer:
This change seems to be a step in the right direction (no pun intended).  So, merging.

Just want to confirm the following cases are handled properly as well:
http://forum.caravelgames.com/viewtopic.php?TopicID=41350&page=0#436048

PR #153, opened 2020-10-09
Pushing player doubles and characters will now activate tokens the sa…

EvidentlyCube:
…me way their movement would

[relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=39223)

### DIFF
--- a/DRODLib/Character.cpp
+++ b/DRODLib/Character.cpp
@@ -5147,6 +5147,12 @@ void CCharacter::PushInDirection(int dx, int dy, bool bStun, CCueEvents &CueEven
 		this->wSwordMovement = nGetO(dx,dy);
 		pGame->ProcessArmedMonsterWeapon(this, CueEvents);
 	}
+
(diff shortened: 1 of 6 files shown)


EvidentlyCube:
@mrimer  You might want to check the thread before merging, since a discussion has emerged.

mrimer:
@EvidentlyCube , thanks for the PR!  I think the code will work as intended, and I appreciate you adding the test cases.

(Tangent: Not sure whether DRODLibTests/DRODLibTest.2019.vcxproj.filters needs to be rebased...)

I'll chime in on the forum topic.  Seems you've been having an extended conversation here, but let's let this PR simmer a bit longer before making a decision.

Hypexion:
I think this PR has simmered long enough, and it's time to close it without merging. No strong support ever emerged for changing it, and there are published rooms that require this behaviour.

mrimer:
Agreed. Closing.

PR #156, opened 2020-10-09
Memory usage optimizations

EvidentlyCube:
Made CRoom::DeadMonsters list be cleared every turn of monsters that are not needed anymore

And dead/ended invisible scripts are also removed from the room.

Previously DeadMonsters was only cleared once upon room object destruction. This caused issues with rooms that do some really heavy scripting and create a lot of temporary characters. Now dead monsters are deleted every turn, except the ones which must persist for longer because they might be processed by frontend for longer. Notably it seems the only case is Speech.

I've carefully inspected every usage of CueEvents private data and failed to find any other situation where a monster attached to an event is stored for future use except speech. Right now the characters used in speech are not 'released' until the room is deleted which shouldn't be a big deal. Even in the worst case I don't foresee more than a few hundred characters being kept like that

If I wasn't careful enough this may cause problems with accessing deleted monsters somewhere in frontend but I have high hopes it'll be fine.

----

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=39045)
[And another](http://forum.caravelgames.com/viewtopic.php?TopicID=41020&page=0#405452)
@mrimer You know the code base best, do you foresee this causing issues somewhere else?

### DIFF
--- a/DRODLib/Character.cpp
+++ b/DRODLib/Character.cpp
@@ -2286,6 +2286,9 @@ void CCharacter::Process(
 				//Deliver speech dialog.
 				if (!command.pSpeech)
 					break; //robustness check
+				//We set this flag to false because this character will be used for as long as the speech
+				//is not finished, so we don't want this object to be deleted next turn
+				this->bSafeToDelete = false;
 				CFiredCharacterCommand *pSpeech = new CFiredCharacterCommand(this, &command,
 					pGame->wTurnNo, this->dwScriptID, this->wCurrentCommandIndex);
 				pSpeech->text = pGame->ExpandText(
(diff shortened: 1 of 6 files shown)


DRODLib/Character.cpp
```diff
@@ -2421,6 +2424,8 @@ void CCharacter::Process(
 								}
 							}
 						}
+						else if (IsAlive())
```
mrimer, line 2427: Suggest a minor refactor to make this easier to grok:

Back at new line 2402, how about:
```
if (IsAlive()) {
  if (!this->bVisible) {
     room.KillInvisibleCharacter(this);
  } else {
     ...everything else...
  }
}
```

DRODLib/DbRooms.cpp
```diff
@@ -8217,16 +8234,38 @@ void CDbRoom::RemoveMonsterFromLayer(CMonster *pMonster)
...
 
 //*****************************************************************************
-void CDbRoom::ClearDeadMonsters()
+void CDbRoom::ClearDeadMonsters(
+	bool bOnlySafe) //(in) WHen true will only delete monsters that are marked as safe to delete
 //Frees memory and resets members for dead monster list.
 {
-	for (list<CMonster *>::const_iterator iSeek = this->DeadMonsters.begin();
+	if (bOnlySafe) {
+		// Not the most elegant solution but I think the as long as the loop's logic is this simple
```
mrimer, line 8242: the as -> that as

mrimer (approved):
Good change!  I love memory optimizations. 👍 

I think this change should be safe.

Made a couple stylistic suggestions.  Can merge without these if you like; just let me know.

EvidentlyCube:
Updated PR

mrimer (approved):

PR #158, opened 2020-10-11
Synchronized Image Overlay effects + multiprocessor compilation

EvidentlyCube:
Added static fields to `CScreen`. One of them is used to replace calls to SDL_Ticks() to ensure effects play synchronized, rather than desyncing in edge cases when ticks crosses a threshold to the next millisecond. Added two more variables which I will be using in future code

Also enabled multiprocessor compilation and incremental linking because they speed things up a lot.

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=44754)


mrimer (approved):

mrimer:
Hey, awesome!  Your fix is just as I envisioned.  Static usage works nicely here in this case.

Didn't know multi-threaded compilation was a thing now.  Excited to try it out!  I'll port this config option usage to rpg, if you haven't done that in another PR yet.

PR #159, opened 2020-10-12
New debug drawing options + gentryii chains correctly update their pa…

EvidentlyCube:
…thmaps now

The old mechanism for detecting for which tiles to update pathmaps had the unfortunate issue of not correctly handling a situation where two diagonally adjacent chain links move in the same orthogonal direction. The new mechanism is much greedier and might cause some duplicates, but for the most part it should be fine, since duplicates are discarded in CCoordSet

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=43499&page=0#436084)

### DIFF
--- a/DROD/RoomWidget.cpp
+++ b/DROD/RoomWidget.cpp
@@ -4488,6 +4488,18 @@ void CRoomWidget::Paint(
 		if (!bPlayerIsAlive)
 			DrawMonsterKillingPlayer(pDestSurface);
 
+		//X. Debug draws, comment out when you want them to be visible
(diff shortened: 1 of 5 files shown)


DROD/RoomWidget.cpp
```diff
@@ -4488,6 +4488,18 @@ void CRoomWidget::Paint(
 		if (!bPlayerIsAlive)
 			DrawMonsterKillingPlayer(pDestSurface);
 
+		//X. Debug draws, comment out when you want them to be visible
```
mrimer, line 4491: comment out -> uncomment

mrimer (approved):

mrimer (approved):

PR #162, opened 2020-10-12
Changed order of drawing room so that dark/light tiles are drawn on t…

EvidentlyCube:
…op of water, stepping stones and shallow water

@mrimer I don't know why the code was specifically saying to draw darkness before shallow water and stepping stones and water, because it doesn't seem to make sense to me, but perhaps there is a reason I just don't understand (maybe the reflection of the sky was supposed to be brighter than the rest of the tile?)

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=41929)

### DIFF
--- a/DROD/RoomWidget.cpp
+++ b/DROD/RoomWidget.cpp
@@ -4010,14 +4010,7 @@ OLayerDone:
 										dest, pDestSurface, 128);
 						}
 
-						//4. Add dark+light to o-layer before sky reflection is rendered.
-						if (bAddLight)
-						{
-							AddLightInterp(pDestSurface, wX, wY, psL, fDark);
-							bAddLightLayers = true;
-						}
-
-						//5. Draw sky/water bottom.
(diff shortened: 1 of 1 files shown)


DROD/RoomWidget.cpp
```diff
@@ -4122,6 +4115,13 @@ void CRoomWidget::RenderRoom(
 							const UINT wTileImageNo = CalcTileImageForWater(this->pRoom, wX, wY, T_THINICE);
 							DrawTransparentRoomTile(wTileImageNo,THIN_ICE_OPACITY);
 						}
+
+						//6. Add dark+light to o-layer after everything else is ddrawn.
```
mrimer, line 4119: ddrawn --> drawn

mrimer (approved):

mrimer:
@EvidentlyCube , IIRC, TFMurphy implemented the shallow water feature and rendering logic.  This is ostensibly a rendering detail he missed.

PR #172, opened 2020-10-13
RPG: Allow editing of hold variables in Global Variable Editor

Hypexion:
When F4 is pressed during playtesting, a list of all game variables is displayed with their values. From this list, the values can be edited.

This PR extends the list to include hold variables, allowed them to be viewed and editing, both with integer and string values. Related thread: http://forum.caravelgames.com/viewtopic.php?TopicID=41659


mrimer (approved):
Nice addition!  Thank you.

PR #174, opened 2020-10-14
Editor ui issues

EvidentlyCube:
- It is no longer possible to change map's display type for a hold you don't own
- It is no longer possible to change map image for a hold you don't own
- When in EditSelectScreen you change from a hold that has world map to one that is not editable, map buttons become disabled now
- It's no longer possible to change 'Show Exit level' for a hold you don't own
- UI elements in EditSelectScreen will no longer be redrawn on top of different elements when changing tabs and focus is on the element outside the tabbed interface

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=42247)

### DIFF
--- a/FrontEndLib/Widget.cpp
+++ b/FrontEndLib/Widget.cpp
@@ -685,6 +685,20 @@ const
 }
 
 //***************************************************************************
+void CWidget::Show(const bool bFlag)
+{
+	this->bIsVisible = bFlag;
+	if (!bFlag)
+		this->Hide(false);
+}
+
+//***************************************************************************
+void CWidget::Hide(const bool /*bPaint*/)
+{
+	this->bIsVisible = false;
+}
+
+//***************************************************************************
 void CWidget::HideChildren()
 //Hides all children of this widget, but not the widget itself.  If you wish
 //to hide both, simply call Hide().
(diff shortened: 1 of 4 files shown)


FrontEndLib/Widget.h
```diff
@@ -372,7 +372,7 @@ class CWidget : public CAttachableObject
 	UINT        GetW() const {return this->w;}
 	UINT        GetH() const {return this->h;}
 
-	virtual void   Hide(const bool /*bPaint*/ = true) {this->bIsVisible = false;}
+	virtual void   Hide(const bool /*bPaint*/ = true);
```
mrimer, line 375: Remove commenting around bPaint here.

FrontEndLib/Widget.cpp
```diff
@@ -684,6 +684,20 @@ const
 	ChildContainerRect.y += this->nChildrenScrollOffsetY;
 }
 
+//***************************************************************************
+void CWidget::Show(const bool bFlag)
+{
+	this->bIsVisible = bFlag;
+	if (!bFlag)
+		this->Hide(false);
```
mrimer, line 692: remove `this->` ?

mrimer (approved):
Thank you!  Looks good.

Couple stylistic suggestions.

FrontEndLib/Widget.h
EvidentlyCube: Why was that commenting there in the first place? I've always been curious whenever I've encountered it. Was it just to stress that the default implementation does not use the argument?

EvidentlyCube:
PR updated

mrimer (approved):

FrontEndLib/Widget.h
mrimer: > Why was that commenting there in the first place? I've always been curious whenever I've encountered it. Was it just to stress that the default implementation does not use the argument?

@EvidentlyCube , great question.  This is actually to suppress a compiler warning.  Any time a method (virtual, in this case) has named arguments that aren't referenced in the body, the compiler may complain about an unused variable.  The way to suppress the variable is to not name the variable in the function signature that's alongside the definition.

PR #179, opened 2020-10-14
Serpent that senses invisible player via a brain smelling the player…

EvidentlyCube:
…, but having no pathmap-path to the player will move like regular snake would and target the player, rather than being confused and acting as if the target was not in the room

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=41313)

### DIFF
--- a/DRODLib/Serpent.cpp
+++ b/DRODLib/Serpent.cpp
@@ -421,7 +421,7 @@ const
 	// If target is ahead or behind the serpent, keep moving toward it.
 	// Otherwise, switch between favoring horizontal or vertical movement
 	// every five turns.
-	if (CanFindSwordsman()) 
+	if (this->pCurrentGame->bBrainSensesSwordsman || CanFindSwordsman())
 	{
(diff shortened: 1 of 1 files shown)


DRODLib/Serpent.cpp
```diff
@@ -68,7 +68,7 @@ const
 		return true;
 	}
 
-	if (!CanFindSwordsman())
+	if (!this->pCurrentGame->bBrainSensesSwordsman && !CanFindSwordsman())
```
mrimer, line 71: I don't think this change is necessary.  Isn't it redundant with the above if check?

mrimer (changes requested):

DRODLib/Serpent.cpp
EvidentlyCube: Good point, reverted.

EvidentlyCube:
PR Updated

mrimer (approved):

PR #186, opened 2020-10-16
Changing editor tabs or selected item during long monster placement w…

EvidentlyCube:
…ill no longer lose the monster placed so far, but rather just place it as if escape was pressed first

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=40835)

### DIFF
--- a/DROD/EditRoomScreen.cpp
+++ b/DROD/EditRoomScreen.cpp
@@ -7106,6 +7106,16 @@ void CEditRoomScreen::SetSelectedObject(const UINT wObject)
 {
 	if (this->wSelectedObject != wObject)
 	{
+		// Finish long-monster placement, otherwise we will lose our changes...
+		if (this->eState == ES_LONGMONSTER) {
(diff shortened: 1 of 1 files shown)


DROD/EditRoomScreen.cpp
```diff
@@ -7106,6 +7106,17 @@ void CEditRoomScreen::SetSelectedObject(const UINT wObject)
 {
 	if (this->wSelectedObject != wObject)
 	{
+		// Place long monster before changing anything ...
+		if (this->eState == ES_LONGMONSTER) {
+			const bool bIsOldTileSerpent = (this->wSelectedObject == T_SERPENT || this->wSelectedObject == T_SERPENTB || this->wSelectedObject == T_SERPENTG);
```
mrimer, line 7111: Do gentryii also need to be considered?

EvidentlyCube: Nope, because this case covers changing serpent type with mouse wheel/clicking the serpent again in the UI. Gentryii have only one type.

EvidentlyCube: I'll change the comments to be clearer

mrimer (approved):

PR #209, opened 2020-10-18
RPG: Local version of statistic modifier variables for NPCs

Hypexion:
Adds new character variables that are used when calculating item value and monster statistic modifiers. These apply when a character is in a room, or when it is equipped as custom equipment.

This requires changes to how modifiers are calculated in `CCurrentGame` and `CDbRoom`, and the addition of a new enum for tracking modifiers.


drodrpg/DRODLib/PlayerStats.cpp
```diff
@@ -52,7 +52,9 @@ const UINT ScriptVars::predefinedVarMIDs[PredefinedVarCount] = {
 	MID_VarEnemyHP, MID_VarEnemyAtk, MID_VarEnemyDef, MID_VarEnemyGold, MID_VarEnemyXP,
 	MID_VarWeaponATK, MID_VarWeaponDEF, MID_VarWeaponGR,
 	MID_VarArmorATK, MID_VarArmorDEF, MID_VarArmorGR,
-	MID_VarAccessoryATK, MID_VarAccessoryDEF, MID_VarAccessoryGR
+	MID_VarAccessoryATK, MID_VarAccessoryDEF, MID_VarAccessoryGR,
+	MID_VarMyMonsterHPMult, MID_VarMyMonsterATKMult, MID_VarMyMonsterDEFMult, MID_VarMyMonsterGRMult,
```
mrimer, line 56: Do we need MID_VarMyMonsterXPMult here?

drodrpg/Texts/MIDs.h
```diff
@@ -1493,6 +1493,16 @@ enum MID_CONSTANT {
   MID_VarItemATKMult = 1568,
   MID_VarItemDEFMult = 1569,
   MID_VarItemGRMult = 1570,
+  MID_VarMyMonsterHPMult = 1753, //New.
```
mrimer, line 1496: Suggest rebuilding DRODUtil so it no longer appends "//New." to new mids.

mrimer (approved):
Looks great!

Thank you for implementing these vars.

I'll mark this PR with tentative approval, but I do have a question I noted in one spot.

Also, we should document these vars in the appropriate help file.

drodrpg/DRODLib/PlayerStats.cpp
Hypexion: We do! Thanks for catching this.

mrimer (approved):

PR #211, opened 2020-10-18
Fixed Beethro never waking up

EvidentlyCube:
[Relevant post](https://forum.caravelgames.com/viewtopic.php?TopicID=38410&page=0#436572)

### DIFF
--- a/DROD/FaceWidget.cpp
+++ b/DROD/FaceWidget.cpp
@@ -340,10 +340,10 @@ void CFaceWidget::SetReading(const bool bReading)
 }
 
 //****************************************************************************
-void CFaceWidget::SetSleeping()
+void CFaceWidget::SetSleeping(const bool bIsSleeping)
 {
-	facePlayer.bIsSleeping = true;
-	facePlayer.bIsBlinking = true;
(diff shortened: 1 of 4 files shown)


DROD/FaceWidget.cpp
```diff
@@ -340,9 +340,9 @@ void CFaceWidget::SetReading(const bool bReading)
 }
 
 //****************************************************************************
-void CFaceWidget::SetSleeping()
+void CFaceWidget::SetSleeping(const bool bIsSleeping)
 {
-	facePlayer.bIsSleeping = true;
+	facePlayer.bIsSleeping = bIsSleeping;
 	facePlayer.bIsBlinking = true;
```
mrimer, line 346: Is bIsBlinking supposed to be false when bIsSleeping is set to false?

EvidentlyCube: Ultimately it does not matter, since it just means that Beethro will have eyes closed for one more FaceWidget frame, but it'll be cleared for future maintainers if it's also set. Updated PR!

mrimer (approved):

PR #215, opened 2020-10-18
It is now enforced that if anything happens to unconquer a room on th…

EvidentlyCube:
…e same turn player is leaving it (be it stairs or Go to level entrance command) the room will remain conquered

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=41104)

### DIFF
--- a/DRODLib/CurrentGame.cpp
+++ b/DRODLib/CurrentGame.cpp
@@ -402,6 +402,8 @@ void CCurrentGame::Clear(
 
 	this->dwCutScene = 0;
 	this->bWaitedOnHotFloorLastTurn = false;
+	this->bWasRoomConqueredAtTurnStart = false;
+	this->bIsLeavingLevel = false;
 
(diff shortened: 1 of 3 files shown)


DRODLib/CurrentGame.cpp
```diff
@@ -6770,6 +6781,8 @@ void CCurrentGame::SetMembersAfterRoomLoad(
 	this->bHalfTurn = false;
 	this->bWaitedOnHotFloorLastTurn = false;
 	this->conquerTokenTurn = NO_CONQUER_TOKEN_TURN;
+	this->bWasRoomConqueredAtTurnStart = false;
```
mrimer, line 6784: Suggest we also init these new vars in Clear(), just in case.

EvidentlyCube: Good point! Also was not resetting `bIsLeavingLevel` here.

mrimer (approved):

PR #221, opened 2020-10-19
Bridge build fixes

EvidentlyCube:
⚠ Review this after #220 is merged, since this branch depends on changes from that PR.

[Relevant thread 1](http://forum.caravelgames.com/viewtopic.php?TopicID=41060)
[Relevant thread 2](http://forum.caravelgames.com/viewtopic.php?TopicID=34576)

### DIFF
--- a/DRODLib/BuildUtil.cpp
+++ b/DRODLib/BuildUtil.cpp
@@ -68,39 +68,57 @@ bool BuildUtil::bIsValidBuildTile(const UINT wTileNo)
 }
 
 //*****************************************************************************
-void BuildUtil::BuildTilesAt(CDbRoom& room, const UINT tile, UINT px, UINT py, const UINT pw, const UINT ph, const bool bAllowSame, CCueEvents& CueEvents)
+UINT BuildUtil::BuildTilesAt(
+// Builds the provided tile in the room at given area. Area is cropped to fit room size, and if the rectangle falls fully outside
+// the room boundaries the functin does nothing
+	CDbRoom& room,                //(in) Room affected
+	const UINT tile,              //(in) Tile to be built, supports virtual tile IDs
+	UINT px, UINT py,             //(in) Position to build at
+	const UINT pw, const UINT ph, //(in) Build rectangle dimensions (use 0, 0 to build a single tile)
+	const bool bAllowSame,        //(in) Whether building the same tile is treated as a successful build or not
+	CCueEvents& CueEvents)
+// Returns: the number of tiles that were modified
 {
 	if (!bIsValidBuildTile(tile))
-		return;
+		return 0;
 
 	UINT endX = px + pw;
 	UINT endY = py + ph;
 
 	if (!room.CropRegion(px, py, endX, endY))
-		return;
+		return 0;
 
 	const UINT baseTile = bConvertFakeElement(tile);
 
-	for (UINT y = py; y <= endY; ++y) {
+	UINT builtTiles = 0;
+	for (UINT y = py; y <= endY; ++y)
 		for (UINT x = px; x <= endX; ++x)
-		{
-			BuildUtil::BuildAnyTile(room, baseTile, tile, x, y, bAllowSame, CueEvents);
-		}
+			if (BuildUtil::BuildAnyTile(room, baseTile, tile, x, y, bAllowSame, CueEvents))
+				++builtTiles;
+
+	if (TILE_LAYER[baseTile] == LAYER_OPAQUE) {
+		for (UINT y = py; y <= endY; ++y)
+			for (UINT x = px; x <= endX; ++x) {
+				//When o-layer changes, refresh bridge supports.
+				room.bridges.built(x, y, room.GetOSquare(x, y));
+
+			}
 	}
+
+	return builtTiles;
 }
 
 //*****************************************************************************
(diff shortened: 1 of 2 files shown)


DRODLib/BuildUtil.cpp
```diff
@@ -67,39 +68,41 @@ bool BuildUtil::bIsValidBuildTile(const UINT wTileNo)
...
 	UINT endY = py + ph;
 
 	if (!room.CropRegion(px, py, endX, endY))
-		return;
+		return false;
 
 	const UINT baseTile = bConvertFakeElement(tile);
 
-	for (UINT y = py; y <= endY; ++y) {
+	bool bAnyBuildSuccessfull = false;
```
mrimer, line 84: bAnyBuildSuccessfull  -> bAnyBuildSuccessful

DRODLib/BuildUtil.cpp
```diff
@@ -484,10 +487,15 @@ bool BuildUtil::BuildNormalTile(CDbRoom& room, const UINT baseTile, const UINT t
 
 	CueEvents.Add(CID_ObjectBuilt, new CAttachableWrapper<UINT>(baseTile), true);
 	
-	//When o-layer changes, refresh bridge supports.
 	if (wLayer == LAYER_OPAQUE) {
-		const UINT newTile = (bIsShallowWater(baseTile) && bIsSteppingStone(room.GetOSquare(x, y))) ? T_STEP_STONE : baseTile;
-		room.bridges.built(x, y, newTile);
+		// Building/removing tiles that (can) affect weapon sheathing should refresh it immediately
```
mrimer, line 491: Rebase this PR following merge of 220.

mrimer (changes requested):

DRODLib/BuildUtil.h
```diff
@@ -38,7 +38,7 @@ class BuildUtil
 public:
 	static bool bIsValidBuildTile(const UINT wTileNo);
 
-	static void BuildTilesAt(CDbRoom& room, const UINT tile, UINT x, UINT y, const UINT w, const UINT h, const bool bAllowSame, CCueEvents& CueEvents);
+	static bool BuildTilesAt(CDbRoom& room, const UINT tile, UINT x, UINT y, const UINT w, const UINT h, const bool bAllowSame, CCueEvents& CueEvents);
```
mrimer, line 41: Pardon, I'm unclear on how returning a bool here is related to this PR.  Can you clarify for me?

EvidentlyCube: Glad to! Previously `BuildUtil::BuildTileAt()` and `BuildUtil::BuildTilesAt()` (note plurality) were completely separate, duplicating some trivial logic which at that time was fine. Now, because bridge processing is supposed to happen at the end and I wanted to avoid duplicating logic, I've made the former just use the latter internally. But the single-tile one returned a boolean value, the multi-tile had to be updated to return some kind of information.

The bool return is used by Citizens to know if the build was successful and they can continue or if they need to wait.

Another option would be to modify Multi-tile method to instead return an UINT with the number of tiles that were actually updated, which could potentially be a bit more useful in the future.

EvidentlyCube: Actually I'll do the change with UINT and add some docs

EvidentlyCube:
PR Updated

DRODLib/BuildUtil.h
mrimer: I appreciate the explanation.  Okay, makes perfect sense.
FYI, the BuildTilesAt return semantic works slighting differently in RPG (for 1.2.8), but it should be fine to keep as you're implementing it here for 5.1.1, as RPG doesn't have any citizen building.

DRODLib/BuildUtil.cpp
```diff
@@ -68,39 +68,57 @@ bool BuildUtil::bIsValidBuildTile(const UINT wTileNo)
...
+	const UINT tile,              //(in) Tile to be built, supports virtual tile IDs
+	UINT px, UINT py,             //(in) Position to build at
+	const UINT pw, const UINT ph, //(in) Build rectangle dimensions (use 0, 0 to build a single tile)
+	const bool bAllowSame,        //(in) Whether building the same tile is treated as a successful build or not
+	CCueEvents& CueEvents)
+// Returns: the number of tiles that were modified
 {
 	if (!bIsValidBuildTile(tile))
-		return;
+		return false;
```
mrimer, line 83: These returns need to be 0, if I'm understanding correctly.

DRODLib/BuildUtil.cpp
```diff
@@ -68,39 +68,57 @@ bool BuildUtil::bIsValidBuildTile(const UINT wTileNo)
...
-	if (!bIsValidBuildTile(tile))
-		return false;
-
-	if (x >= room.wRoomCols || y >= room.wRoomRows)
-		return false; //build area is completely out of room bounds -- do nothing
-
-	const UINT baseTile = bConvertFakeElement(tile);
-
-	return BuildUtil::BuildAnyTile(room, baseTile, tile, x, y, bAllowSame, CueEvents);
+	return BuildTilesAt(room, tile, x, y, 0, 0, bAllowSame, CueEvents) > 0;
```
mrimer, line 121: Like the clean-up possible here.

mrimer (changes requested):
See comment.

EvidentlyCube:
PR updated again!

mrimer (approved):

PR #234, opened 2020-10-23
refactor: refactor out C++11 language features, remove Windows-only INT type

TomBrouws:


EvidentlyCube (approved):

EvidentlyCube:
Looks good though I have two questions:
 - Namespaced references to enums are C++ feature? Or did you remove them to be consistent with the rest of the codebase?
 - Is there any particular reason why we can't use C++11 features?

TomBrouws:
Ah yes, this PR could use a more in-depth description. I had a conversation with Mike about it via PM, and the gist of it is that I'm currently using gcc 4.9 to compile DROD on my Mac, which was the easiest to get working when I first tried to build for OSX. I ran into problems building the latest code, more specifically these three things:
1. The Windows-only INT type. This should be changed regardless.
2. My compiler errors on the namespaced references. I guess newer compiler versions can handle them better. Mike also sent me this link which provides some context: https://stackoverflow.com/questions/514194/using-enum-inside-types-compiler-warning-c4482-c
3. The C++11 lambda function

I told Mike I could try to upgrade my gcc version to avoid making some of these changes, but he argued preserve the ability to build on earlier compiler versions, which I'm also fine with. (And saves me the hassle of upgrading, which judging by the half hour I spent yesterday, is not as simple as it sounds)

@mrimer, you can surely voice your own opinion better than I can, so feel free to add on this.

mrimer:
@TomBrouws , well said!

As the footprint of newer syntactic features and functionality is so small at this point in time, and the benefit of using them is also very small, I think it makes sense to preserve the ability to build on earlier compiler versions for now instead of breaking with them (for non-compelling reasons up to this point).  If for nothing else, than from keeping you from pulling your hair out.  In my experience, getting everything balanced just so on a Mac has been very tricky and I'd rather you not have to deal with attempting to redo that.

At least that's my opinion.  I'd be interested in hearing if anyone has a different sentiment.

EvidentlyCube:
Ah, okay. I totally don't mind not using C++11 features. I didn't even realize I used them, while I do thing I got a pretty good grasp of C++ for the amount of work I've been doing in it, I am not particularly knowledgable about stuff beyond what's in the repository.

And the `INT`s were a dumb accident, probably was looking at some `UINT` while typing them :D.

PR #235, opened 2020-10-23
Allow multi-use plates to be selected in DROD RPG editor

Hypexion:
Muti-use plates are fully implemented in DROD RPG, as far as I can tell. So the only thing that needs to be done is to make them selectable in the editor by re-enabling object rotation for pressure plates, and setting it up so that one-use plates can't be selected.


mrimer (approved):
No kidding!  Thank you for exposing this functionality.

If the engine fully supports these plates, you're right that we can expose them in 1.2.8.

PR #238, opened 2020-10-26
Reduce memory footprint on export and import to avoid OOM issues.

mrimer:
[Relevant issue](https://forum.caravelgames.com/viewtopic.php?TopicID=38449)

This PR addresses OOM errors when attempting to export and import very large .player files (and possibly also .hold files, etc).

Here's what's in this change, which I welcome your feedback on if you have thoughts/opinions:
* Up till now, when exporting, all the data being compiled into the file is first concatenated into a text string in RAM, which is then zlib-compressed into another buffer in RAM, and then CStretchyBuffer-encoded, then written to a file on disk.  Large player profiles (or maybe holds) can run out of memory (in the 32-bit address space in our Windows build) during export.
* To mitigate this, I've changed the export process to use gzip verbs (instead of zlib format), which allow writing of multiple compressed data segments to a file on disk.  I've inserted this operation into the callbackf() method that's called periodically.  Whenever the string gets large (10MB is the heuristic threshold I chose), the buffer is written to disk and cleared out.  Multiple gzip-compressed chunks may be written in sequence to the file.  Now memory usage stays low during export!
* I removed the "Encode" call, which I'm not sure we really need (it's not really secure anyway).  Note this change enables treating the file written to disk as a standard gzip file, and the XML contents can be viewed by anyone who wishes to (rename it as a .gz file and) unpack the contents.  This is probably not a bad thing, but we could re-introduce the encode/decode step if we really want to.
* Both gzip and zlib-plus-encode formats are still supported for import.  Only the gzip version will be written going forward for exports.

_This format change would require adding code to the server to process new v5.1.1+ compressed file uploads, to check for a gzip-format file header (see DbXML.cpp changes) and uncompress the file from gzip format instead of zlib+encode format in that case._   (You don't need to check for any versioning info; just the file header for gzip id bytes.)

The above changes don't address import memory issues.  I spent a good while looking at the DROD address space, which appears to be highly fragmented for large .dat files (by Metakit, if I'm concluding correctly).  I tried out using a custom memory allocator, which greatly reduces the number of distinct mallocs that take place across the address space for sets and maps.  This helped reduce # of allocations (except Metakit char[] allocations).  In certain cases w/ large .dats, I found I wasn't even able to malloc 80MB of contiguous address space, even though the app itself is only using 100-200MB of RAM.  This is ridiculous, but without swapping out Metakit (which would be a lot of work), very large imports would need to be addressed in another way.  If you have any ideas, feel free to share them.  (I can add the custom stl container memory allocator implementation I used in a future PR if it seems interesting to you; I've not included it here.)

Edit: I made a heuristic update to reduce potential import memory allocation, which fortunately resolves the OOM issue on uncompression with the large .player files and large .dat files I tested with.  (Yay!)

Edit2: I also added a progress indicator to the status text being displayed while demos and saved games are being validated for each hold.

### DIFF
--- a/DRODLib/DbXML.cpp
+++ b/DRODLib/DbXML.cpp
@@ -1136,8 +1136,11 @@ void CDbXML::VerifySavedGames()
 	float fNumRecords;
 	UINT wCount;
 	UINT dwTime;
+	WCHAR temp[10];
+	UINT holdCount = 1;
+	const UINT numHolds = info.localHoldIDs.size();
 	for (CIDSet::const_iterator hold=info.localHoldIDs.begin();
-			hold!=info.localHoldIDs.end(); ++hold)
+			hold!=info.localHoldIDs.end(); ++hold, ++holdCount)
 	{
 		//Process data for one hold.
 		bool bMarkedBroken = false;
@@ -1156,8 +1159,17 @@ void CDbXML::VerifySavedGames()
 			{
 				CDbHold *pHold = db.Holds.GetByID(*hold, true);
 				ASSERT(pHold);
-				wstr = wstrHoldName = (const WCHAR*)pHold->NameText;
+				wstrHoldName = (const WCHAR*)pHold->NameText;
 				delete pHold;
+
+				wstr = wszLeftParen;
+				wstr += _itoW(holdCount, temp, 10);
+				wstr += wszForwardSlash;
+				wstr += _itoW(numHolds, temp, 10);
+				wstr += wszRightParen;
+				wstr += wszSpace;
+
+				wstr += wstrHoldName;
 				wstr += wszColon;
 				wstr += wszCRLF;
 				wstr += db.GetMessageText(MID_VerifyingDemos);
@@ -1246,7 +1258,15 @@ void CDbXML::VerifySavedGames()
 				wstrHoldName = (const WCHAR*)pHold->NameText;
 				delete pHold;
 			}
-			wstr = wstrHoldName;
+
+			wstr = wszLeftParen;
+			wstr += _itoW(holdCount, temp, 10);
+			wstr += wszForwardSlash;
+			wstr += _itoW(numHolds, temp, 10);
+			wstr += wszRightParen;
+			wstr += wszSpace;
+
+			wstr += wstrHoldName;
 			wstr += wszColon;
 			wstr += wszCRLF;
 			wstr += db.GetMessageText(MID_VerifyingSavedGames);
@@ -1404,9 +1424,14 @@ MESSAGE_ID CDbXML::Uncompress(BYTE* buffer, UINT size)
 		switch (err)
 		{
 			case Z_BUF_ERROR:
-				//This wasn't enough memory to decode the data to,
-				//so double the buffer size.
-				s_decodedSize *= 2;
+				//This wasn't enough memory to decode the data to.
+				if (s_decodedSize < 30000000) {
+					//...so double the buffer size.
+					s_decodedSize *= 2;
+				} else {
+					//...now try incrementally larger amounts (to attempt to avoid OOM issues, esp. due to 32-bit address fragmentation).
+					s_decodedSize += 10000000;
+				}
 				delete[] s_decodedBuf;
 				s_decodedBuf = NULL;
 				try {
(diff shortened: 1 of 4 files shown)


DRODLib/DbXML.cpp
```diff
@@ -39,9 +39,10 @@
 #include <BackEndLib/StretchyBuffer.h>
 #include <BackEndLib/Ports.h>
 
-#include <zlib.h>
 #include <cstdio>
 
+const char gzID[] = "\x1f\x8b";
```
EvidentlyCube, line 44: Can you add a comment explaining what it is?

DRODLib/DbXML.cpp
```diff
@@ -225,6 +244,8 @@ void CDbXML::PerformCallback(long val) {
 void CDbXML::PerformCallbackf(float fVal) {
 	if (pCallbackObject)
 		pCallbackObject->Callbackf(fVal);
+
+	const bool successIgnored = CDbXML::streamingOut.flush(10 * 1024*1024); //10 MB
```
EvidentlyCube, line 248: Extract the magic number to a constant at the top of the file perhaps?

EvidentlyCube (approved):
Just two minor things

EvidentlyCube:
I am wondering - right now importing loads the whole file into the memory and then imports it. Is there a reason why it couldn't be solved by streaming the file and decoding/inflating on the fly whenever next bytes are requested?

DRODLib/DbXML.cpp
mrimer: Sure. Done!

DRODLib/DbXML.cpp
mrimer: Good idea. Thank you. Done!

mrimer:
> I am wondering - right now importing loads the whole file into the memory and then imports it. Is there a reason why it couldn't be solved by streaming the file and decoding/inflating on the fly whenever next bytes are requested?

That's the right question to ask.  In theory, this is possible.  I believe it could require a large refactor of the XML parsing code, but maybe it's simpler than that.  It would start around CDbXML::ImportXML(), which performs two passes of the uncompressed data buffer.  The uncompression and the parsing steps would need to be coupled.

The first pass of the XML data scans through all the records and counts how many of each type there are, in order to have Metakit pre-allocate space for them to reduce memory address fragmentation.

The second pass reads, populates, identifies, updates foreign keys, and inserts all the records into the DB.

I think that it may be possible to feed in a part of the uncompressed data in a loop for each of these two passes.  If the parser throws an "incomplete data" error, then uncompress more data, append to whatever remained to parse, and continue parsing.

Hopefully this is unnecessary at this point, but if it is ever required to import a large file, this is probably the simplest method to accomplish it.

mrimer:
Updated to address review comments.

EvidentlyCube:
Looks good! While we're at it, can you copy-paste your description of how import works and put it somewhere in the code where someone trying to understand would stumble upon it? Or expand it if you think it'd be a good idea.

For "where", I think `DBXml.cpp` is a good place, and in each `Export` method in the `Db*` classes you could just add a small note "Export/Import is documented in DBXml.cpp". It would certainly help future maintainers :).

mrimer:
@EvidentlyCube , great idea about documentation.  Definitely want future maintainers to be able to easily understand the code design and flow.

Please reference the core `DbXML::ImportXML` method definition for detailed documentation about the import process.

I'll add some more documentation, but generally export/import have a good amount of explanation already in place in DBXML.cpp.

At this point, I believe @schikore still needs to greenlight this general new approach to encoding the exported data, as it will impact server imports.  Once we get approval, we can merge this PR.

mrimer:
Update: I believe I've devised a way to simplify piecemeal uncompression in the import process to further reduce risk of OOM issues.  I'll be adding more documentation and refactoring to make ImportXML easier to follow.  To avoid potential confusion here, I'll prepare that as a distinct PR that I'll put up once this one is approved and merged.

schikore:
> At this point, I believe @schikore still needs to greenlight this general new approach to encoding the exported data as gzip instead of zlib+encode, as it will impact server imports. Once we get approval, we can merge this PR.

@mrimer Can you provide me a file compressed with this new code so that I can run a quick test on the server to make sure it's fine?  I believe it will be, but would rather check to be  sure.

schikore:
Okay, I've got code on the forum that will determine which format is used and properly decode it.  Looks good to me.

schikore (approved):

PR #239, opened 2020-10-26
Weather dialog sliders will no longer draw with weird artefacts + set…

EvidentlyCube:
…tings are not reset on cancel anymore, which caused widgets from different pages to momentarily draw on the screen, as the screen state is reset every time the screen is opened anyway

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=44845)

### DIFF
--- a/DROD/WeatherDialogWidget.cpp
+++ b/DROD/WeatherDialogWidget.cpp
@@ -272,6 +272,9 @@ void CWeatherDialogWidget::SetWidgetStates()
 {
 	ASSERT(this->pRoom);
 
+	// Draw before setting widgets state, as some of them may cache what's drawn underneath them
+	this->RequestPaint(true);
+
 	COptionButtonWidget *pOptionButton;
 	CSliderWidget *pSliderWidget;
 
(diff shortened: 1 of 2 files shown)


DROD/SettingsScreen.cpp
```diff
@@ -1137,41 +1136,6 @@ void CSettingsScreen::OnSelectChange(
...
-	CSliderWidget *pSliderWidget;
-
-	pSliderWidget = DYN_CAST(CSliderWidget*, CWidget*, GetWidget(TAG_GAMMA));
-	g_pTheDBM->SetGamma(pSliderWidget->GetValue());
-
-	pOptionButton = DYN_CAST(COptionButtonWidget*, CWidget*,
-		GetWidget(TAG_ENABLE_SOUNDEFF));
-	g_pTheSound->EnableSoundEffects(pOptionButton->IsChecked());
-	pSliderWidget = DYN_CAST(CSliderWidget*, CWidget*,	GetWidget(TAG_SOUNDEFF_VOLUME));
-	g_pTheSound->SetSoundEffectsVolume(pSliderWidget->GetValue());
```
mrimer, line 1160: I don't think we can cull this method entirely.  The active graphic and sound settings should be reverted back to their original values, correct?  Sounds like the trick is to do so without redrawing the updated widgets on-screen.

mrimer (changes requested):

DROD/SettingsScreen.cpp
EvidentlyCube: Crap, I checked only a few of the lines randomly and it seemed like they don't save any options and of course that was wrong. Updated with a sensible fix!

mrimer (approved):

PR #244, opened 2020-10-27
Additional NPC behaviors

Hypexion:
A few more behaviors, which probably means all the simple to implement ones are done.

**Activate Pressure Plates** is fairly self-explanatory, I think. It replaces the current type check in `CCharacter::CanPressPressurePlates()` with a behavior flag check. Flying and Seep characters get this behavior by default replicate the old logic. This also required a small change to `CDbRoom::ResetUnloadedPressurePlates()` - it now directly checks if a monster can press plates to determine if it will hold down a plate, rather than checking for flight or being a Seep.

**Push Monsters** exists primarily as a hack to maintain previous game logic. Construct-based characters have the ability to push monsters, but not objects. So to avoid any potenial issues, I've split the behaviors in two. This requires a small change to `CMonster`, to add a `CanPushMonsters()` method, and to `CMonster::Move()` to use it when checking if monsters can push another monster. For all predefined monsters, `CanPushMonsters()` calls `CanPushObjects()` to ensure the current logic is maintained.

**Push to Fall Immunity** stops a monster from being pushed into water or pit tiles if that would kill it. This requires a character-specific `CanPushOntoOTileAt()` method to check if a push is allowed, and a change in `CDbRoom::CanPushMonster()`. (One could argue that some refactoring could be done so that monsters deal with what tiles they can be pushed onto instead of the room.)


Data/Help/1/script.html
```diff
@@ -501,7 +501,7 @@ <h3><u>Character Script Commands</u></h3>
...
 	<li><a name="imp-default-pushable-behavior"><b>Default pushable behavior</b></a>: Resets the pushability of the NPC to the default behavior.</li>
-	<li><a name="imp-pushable-by-body"><b>Pushable by body</b></a>: Makes the NPC pushable by the body of the player, mimics, constructs and human-role characters (if a given entity is able to push mirrors it will also be able to push an NPC with this imperative). Regular, non-NPC guards, stalwarts, slayers and soldiers can't push mirrors and thus won't be able to push an NPC with this imperative. The same goes for all monsters except constructs which can push mirrors.</li>
+	<li><a name="imp-pushable-by-body"><b>Pushable by body</b></a>: Makes the NPC pushable by the body of the player, mimics, constructs and charatcers with the <a href="#beh-push-monsters">Push Monsters</a> behavior. Regular, non-NPC guards, stalwarts, slayers and soldiers can't push mirrors and thus won't be able to push an NPC with this imperative. The same goes for all monsters except constructs which can push mirrors.</li>
```
mrimer, line 504: charatcers -> characters

mrimer (commented):
I like these changes!

mrimer:
Thanks for the fix.
Is this PR ready to merge?

mrimer (approved):

DRODLib/Character.cpp
```diff
@@ -3409,6 +3409,42 @@ bool CCharacter::CanPushObjects() const {
...
+
+//*****************************************************************************
+bool CCharacter::CanPushOntoOTileAt(UINT wX, UINT wY) const
+{
+	UINT wTileNo = this->pCurrentGame->pRoom->GetOSquare(wX, wY);
+
+	if (bIsFloor(wTileNo) || bIsOpenDoor(wTileNo) || bIsPlatform(wTileNo))
+		return true;
+
+	// If the character is immune to "fatal pushes", only allow it to pushed
```
EvidentlyCube, line 3426: to pushed -> to be pushed ?

EvidentlyCube (changes requested):
Just one thing in a comment - any chance you could also add a few tests to cover at least a bunch of generic cases for this?

EvidentlyCube (approved):

PR #245, opened 2020-10-28
Effect pausing

EvidentlyCube:
[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=44824)

Posting as draft since it depends on changes from #228. Will update once it's merged.

PR #228, opened 2020-10-20
drawEffect delta time refactor

EvidentlyCube:
Feel free to review the code of individual effects, I don't think it'll be changing anymore, maybe some specific effects that are broken if any are broken. 

The rest is still WIP.

PR #246, opened 2020-10-28
Image overlay bugs

EvidentlyCube:
- Undoing while a infinitely looping overlay is playing every turn won't duplicate the effect for one turn
 -  Effects now pause correctly when speech log is opened and resume when it's closed. Everything also works correctly with losing and regaining focus.
 - Chat history dialog will also pause effects and audio now"
 -  Finally, dialogs will delegate handling windows focus back to their parent (who should either be a screen or another dialog, which in turn will delegate it all the way back to a screen). This allows having special handling for focus, eg. repausing sound if dialog is open

----

Marked as draft since it depends on changes from #228 and #245.

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=44825)

### DIFF
--- a/DROD/GameScreen.cpp
+++ b/DROD/GameScreen.cpp
@@ -6097,6 +6097,26 @@ void CGameScreen::ShowLockIcon(const bool bShow)
 }
 
 //*****************************************************************************
+void CGameScreen::ShowChatHistory(CEntranceSelectDialogWidget* pBox)
+{
+	g_pTheSound->PauseSounds();
+
+	const Uint32 dwSpeechRemaining = this->dwNextSpeech - SDL_GetTicks();
+	this->bIsDialogDisplayed = true;
+	this->pRoomWidget->SetEffectsFrozen(true);
+
+	CDrodScreen::ShowChatHistory(pBox);
+
+	// We compute remaining speech time and replace it instead of just calculating the time the dialog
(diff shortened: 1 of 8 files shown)


mrimer:
@EvidentlyCube , the other PRs are merged.  This PR can now be rebased.

EvidentlyCube:
@mrimer Ready to review and merge

DROD/GameScreen.cpp
```diff
@@ -6069,6 +6096,26 @@ void CGameScreen::ShowLockIcon(const bool bShow)
 	}
 }
 
+//*****************************************************************************
+void CGameScreen::ShowChatHistory(CEntranceSelectDialogWidget* pBox)
+{
+	g_pTheSound->PauseSounds();
+
+	const Uint32 dwSpeechRemaining = this->dwNextSpeech - SDL_GetTicks();
```
mrimer, line 6104: I note here that if dwNextSpeech == 0, then this unsigned value will translate the negative ts to a large positive.

That seems benign, though, as the value won't be used in this case.

mrimer (approved):

mrimer:
@EvidentlyCube , Really elegant improvements built on your nice effect refactoring.  I like the vision you put forth and executed on!

PR #247, opened 2020-10-29
Bump effect fixes.

EvidentlyCube:
- It will no longer populate the draw surface on update, which caused it to pull incorrect surface. This caused a bug when bumping a killing monster - the bump would just draw whiteish square.

 - It will now regenerate on every draw to remove weird visual artifacts

 - Bumping again Row 0 / Col 0 would produce no effect, this is also fixed

[Relevant thread 1](http://forum.caravelgames.com/viewtopic.php?TopicID=44833)
[Relevant thread 2](http://forum.caravelgames.com/viewtopic.php?TopicID=40737)
[Relevant thread 3](http://forum.caravelgames.com/viewtopic.php?TopicID=41029)

### DIFF
--- a/FrontEndLib/BumpObstacleEffect.cpp
+++ b/FrontEndLib/BumpObstacleEffect.cpp
@@ -76,25 +76,47 @@ CBumpObstacleEffect::~CBumpObstacleEffect()
 //********************************************************************************
 bool CBumpObstacleEffect::Update(const UINT wDeltaTime, const Uint32 dwTimeElapsed)
 {
-	SDL_Rect EraseRect = MAKE_SDL_RECT(0, 0, this->src.w, this->src.h);
-	if (!this->pEraseSurface)
-	{
+	if (!this->pEraseSurface) {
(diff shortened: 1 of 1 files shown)


FrontEndLib/BumpObstacleEffect.cpp
```diff
@@ -73,28 +73,51 @@ CBumpObstacleEffect::~CBumpObstacleEffect()
 		SDL_FreeSurface(this->pEraseSurface);
 }
 
+#include "Screen.h"
```
mrimer, line 76: Move #include to top of file.

mrimer (changes requested):

FrontEndLib/BumpObstacleEffect.cpp
EvidentlyCube: That was used for some debugging earlier, removed.

EvidentlyCube:
Ready to review again

mrimer (approved):
So nice to have these issues resolved!

PR #250, opened 2020-10-29
Aumtlich beams double placement

EvidentlyCube:
Aumtlich beams now render during double placement + further cleanup in rendering logic

Previously placing doubles would have a lot of duplicate logic to, I
presume, improve rendering speed by caching almost the whole screen.
This had a couple of problems:

1. A lot of duplicate logic made it very difficult to understand what is
happening where.
2. It was easy to forget to keep both up to date.

As such I refactored it so that it uses common rendering path

----

Depends on PR #246, posting as draft.

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=33595)

### DIFF
--- a/DROD/AumtlichGazeEffect.cpp
+++ b/DROD/AumtlichGazeEffect.cpp
@@ -68,6 +68,9 @@ void CAumtlichGazeEffect::SharedStateUpdate(const UINT wDeltaTime)
 	static bool bMakeBrighter = false;
 	static Uint32 lastUpdatePresentCount = 0;
 
+	if (wDeltaTime == 0)
+		CAumtlichGazeEffect::brightness = 255;
+
 	// Already updated this frame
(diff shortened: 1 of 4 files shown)


mrimer:
@EvidentlyCube , I think this PR is ready to rebase.

mrimer:
@EvidentlyCube , sorry, my last comment was premature.  Now the PR is ready to rebase. :)

EvidentlyCube:
And rebased, ready to review and merge

mrimer (approved):

mrimer:
@EvidentlyCube , I think the cleanup you applied here makes sense.  We don't need mimic placement logic for RPG, but otherwise what changes made here need to be ported over to RPG?

EvidentlyCube:
@mrimer I don't think any change is truly required, since I only modified DROD-specific files. EffectList still has the `UpdateAndDraw()` which is now used only by RPG but that's fine.

PR #253, opened 2020-10-30
Windows: Command line option to change default player data dir location.

mrimer:
To address Metakit filepath issue not working for Windows Steam users w/ non-ascii chars in their username.

### DIFF
--- a/BackEndLib/Files.cpp
+++ b/BackEndLib/Files.cpp
@@ -96,6 +96,7 @@ WSTRING CFiles::wstrHomePath;
 #ifdef WIN32
 bool CFiles::bDrives[NUMDRIVES];
 bool CFiles::bDrivesChecked = false;
+bool CFiles::bWindowsDataFilesInUserSpecificDir = true;
 #endif
 
 #ifdef USE_XDG_BASEDIR_SPEC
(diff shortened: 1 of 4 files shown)


EvidentlyCube:
I find it a little hard to follow the new logic, but wouldn't it make the most sense to do something like this, in pseudocode:
```
function getDataLocation() {
possibleDataLocations = [];

if (GAMEDIR.isWritable())
    possibleDataLocations.push(GAMEDIR);

possibleDataLocations.push("C:/ProgramData/Caravel");

if (username.hasOnlyAsciiCharacters())
    possibleDataLocations .push("C:/Users/<username>/Documents/Caravel");

foreach(location in possibleDataLocations) {
    if (location.hasDataAlready())
        return location
}

return possibleDataLocations[0]; // Or default to the safest C:/ProgramData
}
```

Or, in other words, first build a list of possible places where the data can be stored. Only include `Users/` directory if username has supported characters. Then take the first one which already has DROD data. If none has any data, then just take the first option (or any specific option).

If there is a risk of cross pollinating data files from steam and non-steam versions, then just use a custom suffix for steam version so that it can never happen.

mrimer:
Thanks for reviewing!

I like your ideas 

I can appreciate the sentiment you're sharing to make things bulletproof.

It's possible there could be file conflicts and permissioning issues with multiple users on the system. Rather than check for incidental guard rails. Like ascii chars, we probably need to confirm that the files can be created and opened in the specified location. If not, try the other location.

This may be more convoluted to implement than we think, however, as there's coupling between initialization routines in Main.cpp, CFiles, and Drodlib.

It's not an extremely compelling change to make, imho, but we could try to do this.

EvidentlyCube:
Oh right, since `C:/ProgramData` and `GameDir` would be shared on different user profiles, bummer. I just think that game should not crash unexplicably and require the user to use command line arguments to play the game if they have diacritics in their username, that's an awful UX. Maybe something like this:

Automatically enable the command-line argument if, at the very beginning of the game starting, it detects that user has unsupported characters in their username? So just trigger the option for those users automatically?

mrimer:
I like it!  I'll do that right now.

PR #268, opened 2020-11-01
Replace UnicodeToASCII with UnicodeToUTF8.

mrimer:
Fixes file access issues with extended ASCII and Unicode characters.
Also fix some compiler warnings.

I cleaned up the Steam user dir logic for Windows in Main.cpp.

Edit: I also threw in changing ASCIIToUnicode to UTF8ToUnicode.


EvidentlyCube (approved):
I admit I mostly just scanned through it because I am not really capable of revieing this propely - nothing caught my eye

DROD/Chat.cpp
```diff
@@ -657,7 +657,7 @@ bool Interface::Parse(Json::Value *pChat)
 		Data *c = new Data;
 		c->bAFK = user.get("AFK", false).asBool();
 		string str = user.get("Name", "").asString();
-		AsciiToUnicode(str.c_str(), c->sender);
```
schikore, line 660: Currently the forum will be sending only ASCII, but as I understand it, this will continue to work just fine, and will work if we ever convert the forum to use UTF8 for usernames and other text.  Correct?

schikore (approved):

DROD/Chat.cpp
mrimer: @schikore Yes, that's exactly right.  These changes should forward-proof us against ever wanting to start using UTF8 in these contexts.

PR #270, opened 2020-11-01
Now all cases should be fixed when leaving the room (via Go To Level …

EvidentlyCube:
…Entrance) on the same turn as the room becomes not conquered. Specific cases fixed:

 - Hitting a beacon in a room that was cleared upon entry

 - Creating monsters in a room that was cleared upon entry and then triggering the script

Minimap no longer turns the room to green when this case happens

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=41104)


mrimer (approved):
I confess this logic is getting a bit convoluted for me to follow.  Not your fault!  It makes these cases work properly now, and I thank you for the PR.  Approved!

PR #271, opened 2020-11-01
Animating widgets will no longer happen while deactivating, if deacti…

EvidentlyCube:
…vating occured as part of key repetition code, preventing any stray draws when not wanted

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=38187)


mrimer (approved):
Nice catch!

PR #274, opened 2020-11-02
Death revamp

EvidentlyCube:
Reworked death animation to no longer cache most of the room in a surface and doing shenanigans with effects and other things which caused a lot of issues with image overlays on lower levels, some effects not playing etc. This also makes the death animations much smoother.

[Relevant thred](http://forum.caravelgames.com/viewtopic.php?TopicID=44801)

----

Essentially that will mean death animation will be slightly more CPU intensive, but it will be smoother and all effects will play and there will be no issues with lower-layer image overlays suddenly appearing on top of other things.

### DIFF
--- a/DROD/GameScreen.cpp
+++ b/DROD/GameScreen.cpp
@@ -3020,9 +3020,10 @@ int CGameScreen::HandleEventsForPlayerDeath(CCueEvents &CueEvents)
 		//Fade to black.
 		if (g_pTheBM->bAlpha && dwStart)
 		{
-			const float remainingFraction = 1 - (dwNow - dwStart) / (float)dwDeathDuration;
+			const float durationFraction = min(1, (dwNow - dwStart) / (float)dwDeathDuration);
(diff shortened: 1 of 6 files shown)


EvidentlyCube:
Turned to draft because there are still a few edge cases with certain environmental effects I need to fix.

mrimer:
@EvidentlyCube , I like the direction you're taking this!

EvidentlyCube:
Additional changes:
 * Everything will redraw every frame during death animation to avoid any issues with effects and fadeout
 * Raindrops/Snowflakes will no longer generate during the time when the game is not animating them, to avoid hundreds/thousands being created and lagging/crashing/looking awful
 * Player, dying required hero & killing monsters are now drawn on top of everything, always, and fade has been moved to draw aftere verything else to remove all weird edge cases

And I believe it should be all working neat and cool. Undrafting it.

mrimer (approved):

PR #283, opened 2020-11-03
Guards with daggers

EvidentlyCube:
- Guards with daggers can now dagger step vulnerable characters regardless of their appearance
 - Also tests were broken by the Ascii -> Utf8 change so I fixed them 
 - Fixed the assertion that kept failing during test teardown since forever
 - Also tests are now  guaranteed to run with their own dats

### DIFF
--- a/DRODLibTests/src/Runner.cpp
+++ b/DRODLibTests/src/Runner.cpp
@@ -107,7 +107,7 @@ void Runner::InitializeDatPath(){
 	wstrDatPathTxt += wstrExtension;
 
 	pErrorLogPath = new char[wstrDatPathTxt.length() * 2];
-	UnicodeToUTF8(wstrDatPathTxt.c_str(), pErrorLogPath);
+	UnicodeToAscii(wstrDatPathTxt.c_str(), pErrorLogPath);
 }
 
 CDbRoom* Runner::GetRoom(){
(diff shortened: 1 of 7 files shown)


DRODLibTests/src/Runner.cpp
```diff
@@ -107,7 +107,7 @@ void Runner::InitializeDatPath(){
 	wstrDatPathTxt += wstrExtension;
 
 	pErrorLogPath = new char[wstrDatPathTxt.length() * 2];
-	UnicodeToUTF8(wstrDatPathTxt.c_str(), pErrorLogPath);
+	UnicodeToAscii(wstrDatPathTxt.c_str(), pErrorLogPath);
```
mrimer, line 110: Thank you for correcting my regression here.

Would you please share how my change broke things?  I would not have expected this change to cause any issues.

mrimer (approved):

DRODLibTests/src/Runner.cpp
EvidentlyCube: `pErrorLogPath` is a an array of `char`s, and there seems to be no overload for `UnicodeToUTF8` that converts the text to `char`s. And I need chars because i use `stat()` which accepts that argument.

PR #286, opened 2020-11-04
Fixed the LastNotice flag being updated when it shouldn't

schikore:
LastNotice is used by the HoldSelectScreen to determine if there have been any hold update notices received since the last refresh.  As such, only HoldSelectScreen should update this value.


schikore:
Bug topic:
https://forum.caravelgames.com/viewtopic.php?TopicID=39781

EvidentlyCube (approved):

mrimer (approved):

mrimer:
Thank you!

PR #292, opened 2020-11-05
fix: various changes to fix compilation on OSX

TomBrouws:


DROD/GameScreen.cpp
```diff
@@ -3020,7 +3020,7 @@ int CGameScreen::HandleEventsForPlayerDeath(CCueEvents &CueEvents)
 		//Fade to black.
 		if (g_pTheBM->bAlpha && dwStart)
 		{
-			const float durationFraction = min(1, (dwNow - dwStart) / (float)dwDeathDuration);
+			const float durationFraction = min((float)1, (dwNow - dwStart) / (float)dwDeathDuration);
```
EvidentlyCube, line 3023: Wouldn't `1.0f` be preferable?

EvidentlyCube (approved):

DROD/GameScreen.cpp
mrimer: Yep, let's do it as 1.0f, as the standard idiom for typing a numeric value in C/C++

DROD/Main.cpp
```diff
@@ -1455,7 +1455,9 @@ bool IsAppAlreadyRunning()
 	tmp += dotpid;
 	const UINT lflen = tmp.length();
 	if (!(lockfile = new char[lflen + 1])) return true;
-	UnicodeToUTF8(tmp, lockfile);
+	string lockfileString;
```
mrimer, line 1458: Thanks for this fix.  I agree this will work.  However, we should generally shy away from using char[] buffers for string management in the codebase.  Whenever one of these causes us an issue, we should probably refactor it to be a std::string so we don't need to include code for the memory management, asserting it's non-null wherever it's used, and so forth.  I'm certain I'm to blame for introducing many of these, but as I look back, I acknowledge they're not optimal and we should replace their usage.

FrontEndLib/Sound.cpp
```diff
@@ -985,7 +985,9 @@ SOUNDSTREAM* CSound::LoadSongStream(const WSTRING& wstrSongFilepath, UINT mode)
 	WideCharToMultiByte(CP_ACP, 0, wstrSongFilepath.c_str(), -1, sANSI, 512, NULL, NULL);
 #else
 	char sANSI[wstrSongFilepath.length() + 1];
-	UnicodeToUTF8(wstrSongFilepath.c_str(), sANSI);
+	string sANSIString;
```
mrimer, line 988: Same here.  Let's just replace the sANSI char[] buffer with a string so we don't need to go through these code contortions.

FrontEndLib/Sound.cpp
```diff
@@ -985,7 +985,9 @@ SOUNDSTREAM* CSound::LoadSongStream(const WSTRING& wstrSongFilepath, UINT mode)
 	WideCharToMultiByte(CP_ACP, 0, wstrSongFilepath.c_str(), -1, sANSI, 512, NULL, NULL);
```
mrimer, line 985: I can't recall why we're using WideCharToMultiByte here anymore.  We should replace it.

Not your fault, of course, but I see this function has multiple risks:
https://docs.microsoft.com/en-us/windows/win32/api/stringapiset/nf-stringapiset-widechartomultibyte

While we're improving management of sANSI from a char[] (see my remark below), we should switch sANSI to a string for Windows usage also and use UnicodeToUTF8 to convert wstrSongFilepath to a UTF8 string that FMOD can use to load files.  (Heh...this reminds me of the string conversion issue with the Metakit file loading problem...)

@EvidentlyCube  (same in catch.cpp, probably, to avoid requiring a Windows-specific function for command line arg conversion.)

mrimer (changes requested):

DROD/Main.cpp
TomBrouws: Hmm yes that sounds logical, but in this case a lot of file-system functions are called which all require a char[] buffer. Doesn't it make more sense to leave it as is in this specific case?

FrontEndLib/Sound.cpp
TomBrouws: I can try to make this change.

DROD/GameScreen.cpp
TomBrouws: Sure!

DROD/Main.cpp
mrimer: Good question.  What I'm recommending to do here (and elsewhere, where we allocate char[] buffers) is remove the `lockfile` variable, remove
`strcpy(lockfile, lockfileString.c_str());`
and replace `FILE *fp = fopen(lockfile, "r");` with `FILE *fp = fopen(lockfileString.c_str(), "r");`
This makes more sense to me in order to avoid the superfluous char buffer allocation, since the buffer is already internally allocated inside the std::string var.

mrimer: I'm going to approve and merge this PR and apply my recommended improvements on my end.

mrimer (approved):

PR #293, opened 2020-11-05
Refactor logic for checking behaviors

Hypexion:
This compacts the checks for character behaviors into a single function. Places that used to check `behaviorFlags` now use the `HasBehavior()` function. This should makes things a little more readable, and means if things outside of the character class need to check for behaviors, they can.

There's also a bit of tidying up, including plugging `CanPressPressurePlates()` into a place it should be used.

### DIFF
--- a/DRODLib/Character.h
+++ b/DRODLib/Character.h
@@ -118,7 +118,7 @@ public:
 	virtual UINT   GetIdentity() const {return this->wIdentity;}
 	virtual UINT   GetResolvedIdentity() const;
 	UINT           GetNextSpeechID();
-	bool           HasBehavior(ScriptFlag::Behavior behavior) const { return behaviorFlags.count(behavior); };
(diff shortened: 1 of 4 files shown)


EvidentlyCube (approved):
Love it!

DRODLib/Character.h
```diff
@@ -118,6 +118,7 @@ class CCharacter : public CPlayerDouble
 	virtual UINT   GetIdentity() const {return this->wIdentity;}
 	virtual UINT   GetResolvedIdentity() const;
 	UINT           GetNextSpeechID();
+	bool           HasBehavior(ScriptFlag::Behavior behavior) const { return behaviorFlags.count(behavior); };
```
mrimer, line 121: Suggest keeping the explicit boolean comparison `== 1` here to avoid compiler warnings in some settings regarding the implicit type casting of an integer to a bool.

mrimer (changes requested):
Great refactoring!  One change is needed.

mrimer (approved):

PR #294, opened 2020-11-06
Clone character's sword will no longer flicker under low lighting con…

EvidentlyCube:
…ditions

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=44935&page=0#437374)


mrimer (approved):
Good catch!

PR #295, opened 2020-11-06
Clone movement order will now be preserved when swapping together wit…

EvidentlyCube:
…h movement order of ALL other monsters + swapping clones with tabs will now work correctly if there are non-clones in between (characters with same processing sequence or other doubles) + swapping with tab now works at all

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=44934)

### DIFF
--- a/DRODLib/DbRooms.cpp
+++ b/DRODLib/DbRooms.cpp
@@ -2930,12 +2930,14 @@ bool CDbRoom::DoesSquareContainTeleportationObstacle(
 CMonster* CDbRoom::FindNextClone()
 //Returns: pointer to next monster of type M_CLONE, else NULL if none
 {
-	CClone* pFirstClone = DYN_CAST(CClone*, CMonster*, GetMonsterOfType(M_CLONE));
+	CClone* pFirstClone = DYN_CAST(CClone*, CMonster*, GetMonsterOfType(M_CLONE, true));
 	CMonster* pMonster = pFirstClone;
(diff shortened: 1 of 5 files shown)


DRODLib/DbRooms.cpp
```diff
@@ -3592,7 +3594,9 @@ CMonster* CDbRoom::GetMonsterOfType(const UINT wType) const
 			CCharacter *pCharacter = DYN_CAST(CCharacter*, CMonster*, pMonster);
 			if (pCharacter->IsVisible()) //Character must be in room to count.
 			{
-				if (wType == M_CHARACTER || pCharacter->GetIdentity() == wType)
+				if (wType == M_CHARACTER)
+					return pMonster;
+				else if (pCharacter->GetIdentity() == wType && !bIgnoreCharacterAppearance)
```
mrimer, line 3599: Technically don't need the `else` here.  I think it's easier to read without it, but YMMV.

mrimer (approved):

DRODLib/DbRooms.cpp
EvidentlyCube: No disagreement here but since you merged it it's a little bit too late to change :D.

mrimer: Yeah, no worries.  But!  It's never too late to change. Not suggesting you need to put up another PR for that change though.

PR #303, opened 2020-11-08
Puzzle mode options

EvidentlyCube:
[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=44932)

Note: I have not moved any texts to MIDs file yet. I'll do it tomorrow, but in the meantime feel free to review the rest of the code.

PR #310, opened 2020-11-09
Added _MyColor variable, identical to what is already in RPG

EvidentlyCube:
Depends on PR #303, will undraft once it's merged.

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=42750)

### DIFF
--- a/DRODUtil/Util3_0.cpp
+++ b/DRODUtil/Util3_0.cpp
@@ -925,6 +925,7 @@ const
 	 dwLastMessageID = 0;
 
 	 //Read file into buffer.
+	 std::set<UINT> usedMids;
 	 CStretchyBuffer buf;
 	 CFiles files;
 	 if (!files.ReadFileIntoBuffer(pwzMIDFilepath, buf)) {
@@ -979,6 +980,11 @@ const
 		  //Add element.
 		  string strName = szName;
 		  UINT dwValue = convertToUINT(szValue);
+		  if (usedMids.find(dwValue) != usedMids.end()) {
+			  printf("Duplicate MID value found in key '%s' (value %u)", strName.c_str(), dwValue);
+			  throw false;
+		  }
+		  usedMids.insert(dwValue);
 		  AssignedMIDs[strName] = dwValue;
 
 		  //If this is the largest MID value, remember it.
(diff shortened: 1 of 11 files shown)


mrimer:
I've merged #303

DRODUtil/Util3_0.cpp
```diff
@@ -979,6 +980,11 @@ const
 		  //Add element.
 		  string strName = szName;
 		  UINT dwValue = convertToUINT(szValue);
+		  if (std::find(usedMids.begin(), usedMids.end(), dwValue) != usedMids.end()) {
```
mrimer, line 983: I like this addition as a sanity check.

Suggest using `set<UINT>` instead of `vector<UINT>`, as only one instance of each value is maintained in the data structure.
This reduces algorithmic cost for this loop from N^2 to N log N.

This check can then be simplified to `if (usedMids.count(dwValue) != 0) { ...`
(which has log N instead of N search time).

Data/Help/1/script.html
```diff
@@ -590,9 +588,18 @@ <h3><u>Character Script Commands</u></h3>
...
+You may select these in variable commands to examine or set certain game states.</p>
 <ul>
 	<li><b>_MyWeapon</b> - Sets the weapon wielded by the NPC.</li>
+	<li>
+	<b>_MyColor</b> - Additional tint applied to the character. When the value is -1 (default) no color is applied, otherwise the format is a 6-digit number RRGGBB, where RR is the value for Red tint, GG for Green and BB for Blue. Value of 50 for given color is normal, lesser than 50 will reduce the color, larger than 50 will add the color. Some examples:
+	<ul>
+		<li><b>505050</b> - No color change</li>
+		<li><b>FFFFFF</b> - Twice as bright</li>
+		<li><b>005000</b> - Red and green will be removed</li>
+		<li><b>FF5050</b> - WIll be more red than before</li>
```
mrimer, line 600: WIll --> Will

mrimer (changes requested):

DRODUtil/Util3_0.cpp
EvidentlyCube: I tried but it threw compilation errors I couldn't understand - I'll try again.

mrimer (approved):

PR #314, opened 2020-11-10
Unscramble DROD RPG 1.3 Message Ids

Hypexion:
At some point in development, the MIDs for DROD RPG were... rearranged. This means they no longer point to the correct strings in pre-existing .dat files, which is not good. This PR should put all the old MIDs back to where they should be. However, MIDs added during 1.3 development may have been re-assigned. As this version is not yet released, this shouldn't lead to issues.


mrimer (approved):

mrimer:
Thank you!

Issue #297, opened 2020-11-06
Imperative Vulnerable and Invulnerable have no logic


EvidentlyCube:
It appears in 5.2 there is no longer any logic for setting imperative Vulnerable and Invulnerable or I can't find it. If it's indeed not there then it'll certainly not be backwards compatible.

@Hypexion Tagging so you can take a look.

EvidentlyCube:
Confirmed: A wubba character with Imperative Vulnerable cannot be killed in 5.2 but happily dies in 5.1 to a sword stab.

Hypexion:
> 
> 
> Confirmed: A wubba character with Imperative Vulnerable cannot be killed in 5.2 but happily dies in 5.1 to a sword stab.

This happens because in 5.2, Wubba characters get all the weapon immunity behaviors. Check if this still happens after turning them all off.

Hypexion:
This issue is actually a side effect of #177. It changes the initialisation logic of some characters so that they get a bunch of behaviors instead of the invunerable imperative.

Turns out this breaks backwards compatibility for two reasons:

1. These behaviors aren't removed when you make an NPC vunerable, so they keep being hard to kill when they shouldn't
2. Having the invunerable imperative also conveys immunity to hot tiles and firetraps. #177 removes this intentionally, but this could lead to breaks, especially given how long hot tiles have been in the game.

EvidentlyCube:
I added some test cases that can be used to check if things work the same in 5.2 as in 5.1.1 - https://github.com/CaravelGames/drod/pull/302

My answer to the question posed in the tile is [here](https://github.com/CaravelGames/drod/blob/master/DRODLib/Character.cpp#L2550).

So, my opinion is: we should ditch any existing logic for vulnerable and invulnerable imperative, and just replace it boldly with behaviors, to simplify and granularize properties rather than depend on multiple layers of options which can interact in counter-intuitive ways.

mrimer:
I admit I'm not following the conversation completely, but we need to maintain backwards compatibility.

That is, we can't break pre-existing holds unless we're going to provide a mechanism for transparently updating them to work automagically with the 5.2 logic.

mrimer:
@Hypexion  @EvidentlyCube  We can close this issue now, correct?

EvidentlyCube:
We can, closing!

PR #322, opened 2020-11-11
Merge TSS 5.2 branch with the latest master

Hypexion:
Getting a few things up to speed, in part to be able to deal with #297. After resolving some minor conflicts, everything builds, and only two tests failed. I expected those failures to happen, so things should be good.


mrimer:
Thank you!

PR #327, opened 2020-11-12
Many new controls can now be remapped and they support modifier keys:

EvidentlyCube:
PRIMARY CHANGES:
 - New I18N namespace for functions related to getting a translated text
 - New type InputKey for storing information about a keymapping, a
 combination of SDL's syskey + modifier bits
 - Settings screen reworked to support the new keymappings
 - Many actions can now be remapped
 - Separated KeymapDialogWidget so that DROD and RPG have their own code
 for it

 OTHER CHANGES:
 - Refactored settings screen to separate widget setup into its own
 methods
 - Setting screen now uses a new font for keymaps, changing spacing
 between keymap widgets
 - Other code for handling player input now also supports InputKey
 can also hold modifier data
 - CDBPackedVars now supports 64 bit signed integers
 - CLabelWidget now supports vertically centered text
 - Added new font that's slightly smaller that is used for showing
 keybindings on settings screen, so that they can be split into two
 lines to be able to fit modifiers together with the keybinding

[Relevant thread](https://forum.caravelgames.com/viewtopic.php?TopicID=44840)


DROD/SettingsScreen.cpp
```diff
@@ -1052,6 +1109,86 @@ void CSettingsScreen::OnClick(const UINT dwTagNo)
...
+
+		// If we are changing a command that does not support modifiers, and provide a key with
+		// modifiers, make sure we don't use the same base key as any of the commands
+		// that have modifiers-variants built-in, eg. ctrl+move
+		if (!DoesCommandUseModifiers(eCommand) && (InputKey)ReadInputKey(newKey) != newKey)
+		{
+			const KeyDefinition *pOverwrittenKey = GetOverwrittenModifiableKey(newKey, eCommand);
+
+			if (pOverwrittenKey) {
+				WSTRING str = L"You can't map to this key because command '";
```
mrimer, line 1155: This text should be converted into a MID (or two) by adding new entries into a file in the Texts dir (plus entries in DbBase.cpp for patching).

DROD/SettingsScreen.cpp
```diff
@@ -1052,6 +1109,86 @@ void CSettingsScreen::OnClick(const UINT dwTagNo)
...
+	if (dwTagNo < TAG_KEY_BUTTON_START || dwTagNo > TAG_KEY_BUTTON_END)
+		return;
+
+	InputKey newKey, currentKey;
+
+	const DCMD eCommand = ButtonTagToDcmd(dwTagNo);
+	const KeyDefinition *keyDefinition = GetKeyDefinition(eCommand);
+	currentKey = this->pCurrentPlayer->Settings.GetVar(keyDefinition->settingName, UNKNOWN_INPUT_KEY);
+
+	if (GetCommandKeyRedefinition(eCommand, currentKey, newKey, !DoesCommandUseModifiers(eCommand)))
```
mrimer, line 1142: For readability, suggest inverting the check here (i.e, prepend "!") and place the else block immediately underneath it.  Then you don't need all the rest of the code indented inside another block below that.

DRODLib/GameConstants.cpp
```diff
@@ -60,32 +60,72 @@ const WCHAR wszVersionReleaseNumber[] = {
...
-		MID_MoveWest, MID_Wait, MID_MoveEast,
-		MID_MoveSouthwest, MID_MoveSouth, MID_MoveSoutheast,
-		MID_SwingClockwise, MID_SwingCounterclockwise,
-		MID_RestartRoom, MID_UndoMove, MID_BattleKey, MID_UseCommandKey, MID_CloneSwitch
-	};
+	const std::unordered_map<DCMD, KeyDefinition*> BuildKeyDefinitions()
+	{
+		std::unordered_map<DCMD, KeyDefinition*> keyDefinitions;
+		
+		keyDefinitions[DCMD_NW] = new KeyDefinition(CMD_NW, "MoveNorthwest", MID_MoveNorthwest, SDLK_KP_7, SDLK_7);
```
mrimer, line 67: Having everything for each definition laid out all in one place is much nicer than what we had before!

DRODLib/GameConstants.h
```diff
@@ -27,12 +27,18 @@
 #ifndef GAMECONSTANTS_H
 #define GAMECONSTANTS_H
 
+#include <map>
+#include <unordered_map>
```
mrimer, line 31: As a rule of thumb, put these std include files below all the other includes, so that the files included lower down don't inadvertently satisfy a dependency they're not including themselves (but should be), by having these expanded into the source file above them. If that makes sense.

Texts/MIDs.h
```diff
@@ -1,5 +1,5 @@
 //MIDs.h
-//Generated by DRODUtil
+//Generated by DRODUTIL at 11/12/20 11:16PM.
```
mrimer, line 2: Revert this line's change.
Re-compile DRODUtil so it doesn't generate this timestamp anymore.  We don't need the timestamp with version control, and it creates merge conflicts for us to resolve.

mrimer (changes requested):
Beautiful work!

I think I followed everything and don't see anything I have questions or concerns about.

Remarked on minor changes to make.

DRODLib/GameConstants.cpp
EvidentlyCube: Also it's less error prone :).

EvidentlyCube:
Pushed fixes

Texts/MIDs.h
```diff
@@ -1,5 +1,5 @@
 //MIDs.h
-//Generated by DRODUtil
+//Generated by DRODUTIL
```
mrimer, line 2: This change is close but not identical to the original line (case difference).  Let's revert this change back to the original line.

Texts/MIDs.h
```diff
@@ -1316,6 +1316,7 @@ enum MID_CONSTANT {
   MID_GetKeyDescription_YesModifiers = 1966,
   MID_DefineAllKeys = 957,
   MID_InvalidCommandKey = 958,
+  MID_OverwrittingMacroKeyError = 1993,
```
mrimer, line 1319: MID_OverwrittingMacroKeyError -> MID_OverwritingMacroKeyError

DROD/SettingsScreen.cpp
```diff
@@ -1139,54 +1138,54 @@ void CSettingsScreen::DoKeyRedefinition(const UINT dwTagNo) {
...
-			}
+	// If we are changing a command that does not support modifiers, and provide a key with
+	// modifiers, make sure we don't use the same base key as any of the commands
+	// that have modifiers-variants built-in, eg. ctrl+move
+	if (!DoesCommandUseModifiers(eCommand) && (InputKey)ReadInputKey(newKey) != newKey)
+	{
+		const KeyDefinition *pOverwrittenKey = GetOverwrittenModifiableKey(newKey, eCommand);
+
+		if (pOverwrittenKey) {
+			WSTRING str = WCSReplace(
```
mrimer, line 1157: I like this approach for interpolating a parameter in the text.

mrimer (changes requested):

EvidentlyCube:
Let's try again!

mrimer (approved):

PR #328, opened 2020-11-13
 RPG: Add global script vars to modify what entity types are spawned. 

mrimer:
For mud/tar/gel and roach queens.


mrimer:
I've merged this, but feel free to share feedback after the fact.

PR #330, opened 2020-11-13
Fix potential behavior compatibility issues

Hypexion:
A lot of scripted character interactions that are now controlled by behaviors used to rely on the character's identity to decide what would happen. As default behaviors are only set when the character is initialized, this could lead to problems with scriptings using the `Set appearance` command.

To fix this, I've factored out the default behavior setting into it's own function, which can be called when a character's identity is changed by `Set appearance`. I've also added a bit to the help to inform users of this.

I've also reverted #177 as part of this, for two reasons. The first is that changing what a character identity is immune to could break rooms, which is bad. The second reason is that imperatives, unlike behaviors, persist across identity changes. Without adding a bunch of special handling for the Vulnerable and Invulerable imperatives, this acts as another way for room breaks to creep in, in a way that could be pretty difficult to spot.


mrimer:
The changes look good.  I'm not clear on the resultant state of the code after this PR is merged.  Are all regressions from pre-5.2 logic fixed?

Hypexion:
> 
> The changes look good. I'm not clear on the resultant state of the code after this PR is merged. Are all regressions from pre-5.2 logic fixed?

Regressions related to the invulnerable imperative should be fixed now, assuming the tests are testing correctly. I don't know if the identity related stuff would actually cause any rooms breaks, but now things should be consistent with how there were before.

Hypexion:
Incidentally since this PR is adding new string it might be best to wait until #327 is merged to make sure things related to that are okay.

mrimer:
> Incidentally since this PR is adding new string it might be best to wait until #327 is merged to make sure things related to that are okay.

Appreciate the guidance.  Will do.

mrimer:
I've merged #327.  This PR is ready for you to rebase.

mrimer (approved):

PR #331, opened 2020-11-14
RPG: Fix regression in smooth lighting animation on moving pit platforms.  Other lighting fixes and improvements.

mrimer:
[Relevant topic](https://forum.caravelgames.com/viewtopic.php?TopicID=44947)

Multiple changes here.  Check out the commit diffs individually for the recommended way to follow the changes.

If this looks good to you, I'll migrate these fixes to 5.1.1.

### DIFF
--- a/FrontEndLib/BitmapManager.h
+++ b/FrontEndLib/BitmapManager.h
@@ -68,6 +68,15 @@ struct CustomTiles
 	UINT wTileStartNo;     //starting number of these custom tiles
 };
 
+struct TileMask
+{
+	TileMask(UINT tile, int xOffset, int yOffset)
+		: tile(tile), xOffset(xOffset), yOffset(yOffset)
+	{ }
+	UINT tile;
+	int xOffset, yOffset; //tile location offset from top-left origin
+};
+
 //DataIDs to find custom tile data that the bitmap manager may use
 typedef std::map<UINT,CustomTiles> CustomTileMap;
 
@@ -134,6 +143,9 @@ public:
 	void        DarkenTileWithMask(const UINT wTIMask, const UINT wXOffset, const UINT wYOffset,
 			const UINT x, const UINT y, const UINT w, const UINT h,
 			SDL_Surface *pDestSurface, const float fLightPercent);
+	void CBitmapManager::DarkenTileWithMultiTileMask(const vector<TileMask>& masks,
+			const UINT x, const UINT y, const UINT w, const UINT h,
+			SDL_Surface* pDestSurface, const float fLightPercent);
 	void        DarkenWithMask(SDL_Surface *pMaskSurface, SDL_Rect src,
 			SDL_Surface *pDestSurface, SDL_Rect dest, const float fLightFactor,
 			const bool bUseColorKeyMask=false);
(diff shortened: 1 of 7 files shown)


EvidentlyCube:
Seems fine, I think I'll have to see it migrated to DROD codebase to better understand all the changes.

EvidentlyCube:
I am reviewing it though! That was just first impression comment!

FrontEndLib/BitmapManager.h
```diff
@@ -68,6 +68,15 @@ struct CustomTiles
 	UINT wTileStartNo;     //starting number of these custom tiles
 };
 
+struct TileMask
+{
+	TileMask(UINT tile, int xOffset, int yOffset)
+		: tile(tile), xOffset(xOffset), yOffset(yOffset)
+	{ }
+	UINT tile;
```
EvidentlyCube, line 76: Maybe `tileImage`? Though looking at the rest of the code, TI's are usually referred to as tiles.

FrontEndLib/BitmapManager.h
```diff
@@ -68,6 +68,15 @@ struct CustomTiles
...
+	int xOffset, yOffset; //tile location offset from top-left origin
```
EvidentlyCube, line 77: Top-left origin meaning top-left of the room widget? surface? screen? tile?

I am guessing tile, can you add a tiny comment at the top of the struct explaining how it's supposed to be used?

EvidentlyCube (approved):
Approving but I'd like a few more comments where I mentioned it. I think I get now how it's supposed to work, though I do worry a little bit that it's a very specific solution, and if we ever end up needing something similar for anything else it'll be a lot of work to make it work. But then again I think it's high time to get 5.1.1 out the door, let's avoid refactor creep.

FrontEndLib/BitmapManager.h
mrimer: Agree this should be renamed to be more descriptive.
I'm renaming it to TweeningTileMask to indicate it's used for tiles in motion.

FrontEndLib/BitmapManager.h
mrimer: Good call out.  I've clarified the comment:
`//relative tile location offset from top-left origin of the area this mask is being applied to`

mrimer:
> I do worry a little bit that it's a very specific solution, and if we ever end up needing something similar for anything else it'll be a lot of work to make it work.

This is a good call out, and I don't disagree.  This was a fair amount of work, and it is a specific solution.  Still, we might find the new routine in CBitmapManager to be something that can be recycled for future complex rendering operations, even if we stop using it for this particular use case.

Fortunately, my solution to fixing the pit darkening is localized to one spot in CRoomWidget, so it would be easy to remove.  If we need to generalize this type of operation (or go with something completely different) in the future, I have no qualms (in principle) about ripping that out and replacing it.

mrimer:
BTW, to check out the light rendering changes, they are easiest to observe if you slow down the key repeat rate and try moving around in a very dark room.  Try also stepping down from a wall or top of a door for z-axis light tweening!  This will also all work when migrated to 5.x.

mrimer:
Ready to review my latest commit.

PR #336, opened 2020-11-17
RPG: Correct errors in clipping logic and light application for sprites moving across room boundary.

mrimer:
[Relevant topic](https://forum.caravelgames.com/viewtopic.php?TopicID=44953)

There were multiple issues with the clipping and lighting logic at the room boundary.  These changes are much easier to see with the larger tiles of RPG.

Once reviewed and approved, I can migrate these changes to 5.x.

### DIFF
--- a/drodrpg/DRODLib/DbRooms.cpp
+++ b/drodrpg/DRODLib/DbRooms.cpp
@@ -4975,22 +4975,28 @@ const
 }
 
 //*****************************************************************************
+void CDbRoom::SnapCoordsOutsideOfRoomToEdge(int& nX, int& nY) const
+{
+	if (nX < 0)
+		nX = 0;
+	else if ((UINT)nX >= this->wRoomCols)
+		nX = this->wRoomCols - 1;
+
+	if (nY < 0)
+		nY = 0;
+	else if ((UINT)nY >= this->wRoomRows)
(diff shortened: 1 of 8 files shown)


drodrpg/DRODLib/DbRooms.cpp
```diff
@@ -4974,42 +4974,40 @@ const
 	return this->pszTParams[ARRAYINDEX(wX,wY)];
 }
 
+//*****************************************************************************
+void CDbRoom::SnapCoordsOutsideOfRoomToEdge(int& nX, int& nY) const
```
EvidentlyCube, line 4978: Clamp would probably be a better word to use, since snap suggests snapping to grid. Maybe `ClampCoordsToRoom`?

EvidentlyCube (changes requested):
Got one comment

drodrpg/DRODLib/DbRooms.cpp
mrimer: That's a good callout.  The room is a grid, but I agree Clamp makes more sense with the logic of the function bounding outside coords to the room edges.  I'll make the change.

mrimer:
I'll start migrating these changes to 5.1.1 now.

EvidentlyCube (approved):

PR #340, opened 2020-11-18
Added some functionality to the buttons on the notices popup

schikore:
This makes the buttons on the notices popup do something:
- The delete button removes the selected notice from the list
- The watch button will do different things depending on the type of notice:
-- Hold release/update - take user to hold select screen with the new/updated hold selected
-- URL - open a browser with the selected URL.  Currently there are no notices that use this, but now we can add them!
-- There is a third type defined in DROD that we also haven't used, one that specifies a room.  The main idea was to have a "your #1 score was beaten" notice, and the button would take you to that room if installed & available.  

Relevant thread: https://forum.caravelgames.com/viewtopic.php?TopicID=39139

### DIFF
--- a/DROD/HoldSelectScreen.cpp
+++ b/DROD/HoldSelectScreen.cpp
@@ -960,6 +960,16 @@ void CHoldSelectScreen::OnBetweenEvents()
 		this->bDownloadList = false;
 
 		this->pSelCNetHold = NULL;
+		PopulateHoldListBox();
+
+		UINT caravelNetSelectHoldId = (UINT)g_pTheNet->GetDownloadHold();
+		if (caravelNetSelectHoldId != 0) {
+			// User got here from a CaravelNet new/update hold notice.  Find the hold and select it.
+			this->pHoldListBoxWidget->SelectItem(caravelNetSelectHoldId);
+			g_pTheNet->SetDownloadHold(0);
+		}
+
+
 		SetHoldDesc();
 		Paint();
 	}
@@ -1617,12 +1627,6 @@ void CHoldSelectScreen::SetHoldFilter()
 				g_pTheSM->GetScreen(SCR_EditSelect));
 		if (pESS->GetPrevHoldID())
 			dwSelectHoldID = pESS->GetPrevHoldID();
-	}
-	UINT caravelNetSelectHoldId = (UINT)g_pTheNet->GetDownloadHold();
-	if (caravelNetSelectHoldId != 0) {
-		// User got here from a CaravelNet new/update hold notice.  Find the hold and select it.
-		dwSelectHoldID = caravelNetSelectHoldId;
-		g_pTheNet->SetDownloadHold(0);
 	}
 
 	this->pHoldListBoxWidget->SelectItem(dwSelectHoldID);
--- a/DROD/HoldSelectScreen.cpp
+++ b/DROD/HoldSelectScreen.cpp
@@ -962,7 +962,7 @@ void CHoldSelectScreen::OnBetweenEvents()
 		this->pSelCNetHold = NULL;
 		PopulateHoldListBox();
 
-		UINT caravelNetSelectHoldId = (UINT)g_pTheNet->GetDownloadHold();
+		UINT caravelNetSelectHoldId = g_pTheNet->GetDownloadHold();
 		if (caravelNetSelectHoldId != 0) {
 			// User got here from a CaravelNet new/update hold notice.  Find the hold and select it.
 			this->pHoldListBoxWidget->SelectItem(caravelNetSelectHoldId);
--- a/DROD/HoldSelectScreen.cpp
+++ b/DROD/HoldSelectScreen.cpp
@@ -958,7 +958,6 @@ void CHoldSelectScreen::OnBetweenEvents()
 		// Done downloading the list.
 		CScreen::HideStatusMessage();
 		this->bDownloadList = false;
-		PopulateHoldListBox();
 
 		this->pSelCNetHold = NULL;
 		SetHoldDesc();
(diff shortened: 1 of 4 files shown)


DROD/HoldSelectScreen.cpp
```diff
@@ -958,7 +958,6 @@ void CHoldSelectScreen::OnBetweenEvents()
 		// Done downloading the list.
 		CScreen::HideStatusMessage();
 		this->bDownloadList = false;
-		PopulateHoldListBox();
```
EvidentlyCube, line 961: How is the hold list populated now that this is removed?

DROD/RoomScreen.h
```diff
@@ -73,6 +73,7 @@ class CRoomScreen : public CDrodScreen
 	void     ShowScroll() {this->bIsScrollVisible = true; PaintScroll();}
 	void     OpenNoticesBox(CRoomWidget *pRoomWidget);
 	void     GrabNewNotices(CRoomWidget *pRoomWidget);
+	const void UpdateNoticesButton();
```
EvidentlyCube, line 76: I recall reading recently that `const void` does nothing but I can't find it now - can someone who knows C++ better than me confirm/deny?

DROD/RoomScreen.cpp
```diff
@@ -427,13 +428,73 @@ void CRoomScreen::AddNoticesDialog()
...
+{
+	const int num = this->notices.size();
+	CButtonWidget* pButton = dynamic_cast<CButtonWidget*>(this->GetWidget(TAG_OPEN_NOTICES));
+	string label;
+	WSTRING wLabel;
+	if (num >= 10) {
+		label = "...";
+	}
+	else {
+		char buffer[5];
```
mrimer, line 488: I know we're just refactoring code location, but while we're at it, let's set buffer's size to 12 to avoid potential overflow issues (even if not realistic, we don't want to open the door to them).

DROD/RoomScreen.h
mrimer: I've never used `const void` before, but we can remove the 'const'.

DRODLib/NetInterface.h
```diff
@@ -221,6 +222,8 @@ class CNetInterface {
 	// Notices
 	virtual void SetLastNotice(const UINT /*dwNoticeID*/) {}
 	virtual void QueryNotices(vector<CNetNotice> &/*notices*/, UINT /*typeMask*/, UINT /*lastId*/) const {}
+	virtual void SetDownloadHold(const int /*dwServerHoldId*/) {}
```
mrimer, line 225: Not having seen the implementation myself yet, is there a reason to use `int` with the new methods instead of `UINT`?

schikore: Ah.  I used int because it *is* a signed int, beta holds have a negative id.  But in other places we do cast this to UINT, so I'll just update it to store it that way.

DROD/HoldSelectScreen.cpp
schikore: PopulateHoldList was being called twice.  This one was superfluous.

DROD/RoomScreen.cpp
schikore: This is in an else. The if condition is "if (num >= 10)".  Five is safe here.

DROD/HoldSelectScreen.cpp
mrimer: I must be being dense. I'm not seeing the second call. Would you point it out to me?

DROD/HoldSelectScreen.cpp
```diff
@@ -1619,6 +1618,13 @@ void CHoldSelectScreen::SetHoldFilter()
 		if (pESS->GetPrevHoldID())
 			dwSelectHoldID = pESS->GetPrevHoldID();
 	}
+	UINT caravelNetSelectHoldId = (UINT)g_pTheNet->GetDownloadHold();
+	if (caravelNetSelectHoldId != 0) {
```
mrimer, line 1622: Can remove this UINT cast now.

DROD/HoldSelectScreen.cpp
schikore: Okay.  I believe this is correct now.  It *does* re-populate the list here, but now selects the hold in question in this function, upon this repopulation.

DROD/HoldSelectScreen.cpp
```diff
@@ -958,9 +958,18 @@ void CHoldSelectScreen::OnBetweenEvents()
 		// Done downloading the list.
 		CScreen::HideStatusMessage();
 		this->bDownloadList = false;
-		PopulateHoldListBox();
 
 		this->pSelCNetHold = NULL;
+		PopulateHoldListBox();
+
+		UINT caravelNetSelectHoldId = (UINT)g_pTheNet->GetDownloadHold();
```
mrimer, line 965: K, thanks, this makes sense.

Can remove the UINT cast here now,

DROD/RoomScreen.cpp
```diff
@@ -427,13 +428,72 @@ void CRoomScreen::AddNoticesDialog()
 void CRoomScreen::OpenNoticesBox(CRoomWidget* pRoomWidget)
 {
 	this->pNoticesDialog->SetBetweenEventsHandler(GetEventHandlerWidget()); //keep updating room effects
-	const UINT returnTag = this->pNoticesDialog->Display(false);
+	UINT returnTag = 0;
```
mrimer, line 431: This is awesome functionality.  I love it!

DROD/HoldSelectScreen.cpp
schikore: Done, missed that comment earlier.

mrimer (approved):

PR #344, opened 2020-11-19
Added three new INI options:

EvidentlyCube:
- Full Screen mode
 - Allowing resizing of the game window
 - Automatically minimizing the game when alt tabbing on full-screen

[Relevant thread](https://forum.caravelgames.com/viewtopic.php?TopicID=44965)

### DIFF
--- a/DROD/Main.cpp
+++ b/DROD/Main.cpp
@@ -472,10 +472,11 @@ int main(int argc, char *argv[])
 			const bool bFullscreen = !CScreen::bAllowWindowed || (!bNoFullscreen && s.GetVar(Settings::Fullscreen, false));
 			if (bFullscreen)
 			{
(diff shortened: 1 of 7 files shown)


mrimer (approved):

vermeeren:
@EvidentlyCube Do you recall there being a reason why the window resizing is disabled by default? I know with SDL1 it cannot scale but with all future releases being SDL2 I think it being on by default will be helpful to the majority of players.

For more context see https://forum.caravelgames.com/viewtopic.php?TopicID=45538. (Of which I am OP.)

EvidentlyCube:
@vermeeren Whatever the reasons were, they were probably bad. Maybe to avoid people accidentally resizing the window and being unhappy about not being able to easily resize it back to maintain 1:1 pixel ratio.

PR #348, opened 2020-11-21
Improved media management options:

EvidentlyCube:
Functional:
 - Deleted media files are not immediately removed from the list if you hit "Delete", they are marked as "(pending deletion)" first
 - Added a new button for undeleting a media file that is pending deletion
 - Added a new button for replacing a media file with a different one. This updates the file name to a new one. If you select a file with a name that is already used, it'll refuse to do so.
 - When trying to add a new file and it shares the name with a file that already exists, it'll ask you if you want to replace it instead. It will also unmark it from deletion if required
 - All places where media files are used in DROD now support all the above options: selecting image by F9, selecting overlay image, adding sound to speech, ambient sound, video, image overlay
 - Fixed a bug where replacing a speech command that had a sound with a different command would almost always result in a crash

INTERNAL:
 - Added new dialog for managing media files

TESTER NOTES:
 - Check all the media management dialog options in evey place where they appear, just in case
 - Thoroughly check what happens with replacing the speech command that has a sound effect attached in different scenarios: switching to another command, switching to another command that also uses sound, copying the command, cut-and-pasting, copy/cut between a character and a character's default script, or between two different default scripts, or between rooms, or between holds.

----

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=44359)


drod/SelectMediaDialogWidget.cpp
```diff
@@ -0,0 +1,574 @@
...
+			wstrImportPath,
+			wstrSelectedFile,
+			fileDialogHeaderMID,
+			false,
+			allowedExtensionsFlags
+		);
+		if (dwSelectFileResult != TAG_OK)
+			break;
+
+		//3. Update the path in player settings, so next time dialog comes up it will have the same path.
```
mrimer, line 396: 3. -> 2.

drod/SelectMediaDialogWidget.cpp
```diff
@@ -0,0 +1,574 @@
...
+				if (!ValidateVideo(buffer)) {
+					g_pTheDialogs->ShowOkMessage(MID_FileNotValid);
+					continue;
+				}
+				g_pTheDialogs->HideStatusMessage();
+				wDataFormat = DATA_THEORA;
+				break;
+		}
+
+		// 5. Import the file
```
mrimer, line 458: 5. -> 6.

mrimer (changes requested):
Great refactor and also new functionality!

Couple minor changes suggested.

EvidentlyCube:
Updated the PR!

EvidentlyCube:
Added two files that I forgot to push earlier.

mrimer (approved):

PR #350, opened 2020-11-23
OSX: Replace a char[] with string

mrimer:


DROD/Main.cpp
```diff
@@ -1470,14 +1467,10 @@ bool IsAppAlreadyRunning()
 	tmp += wszSlash;
 	tmp += wszDROD;
 	tmp += dotpid;
-	const UINT lflen = tmp.length();
-	if (!(lockfile = new char[lflen + 1])) return true;
```
mrimer, line 1474: @TomBrouws , note that if there is any expansion in the number of characters due to the UCS-16 to UTF8 conversion, the char buffer size here would prove insufficient and could result in a crash.

mrimer:
@TomBrouws , feel free to make any further changes on top of these, as needed.

TomBrouws:
@mrimer Thanks for making the changes, this makes sense and looks good to me!

PR #353, opened 2020-11-23
Added a dynamic filtering to the listbox widget and enabled it on all…

EvidentlyCube:
… lists which tend to have a lot of values and could use additional filtering.

FUNCTIONAL:
 - Fixed a bug where using shift+tab to navigate back through the focus list would never focus the first element -> for example in the add command dialog you could never get back to the list of commands by using shift+tab
 - Added a new filtering to some of the lists which works like this:
   - Type letters and space to filter the list
   - It will limit the items in the list only to those that match every word you type
 - Enabled the filtering on:
   - Add command dialog: list of commands and most of the lists in there
   - Customize character dialog: character list
   - Character management dialog: custom characters list, type list
   - Build Levels screen: list of holds
   - Hold selection screen: list of hold

INTERNAL:
 - Added some fancy functions to WChar.cpp that are used to implement the filtering
 - Select/Unselect on CFocusWidget are virtual so you can things happen on losing focus
 - CWidget has a way to prevent event bubbling, currently only works KeyDown but can be easily expanded

NOTES FOR TESTERS:
 - Here are some issues I've found/introduced during development and fixed. It's best to double check them, and you can test all of them on the command list dialog.:
   a) Select some option -> Filter so that it's not available -> Up/Down arrows wouldn't select things
   b) (Add Command dialog) Select some option -> Filter so that it's not available -> widgets for the selected option would not be available
   c) Scrollbar was not resized
   d) The last item in the list was not available in the filtering if there were more things in the list than can fit
   e) Filter -> Select an item -> remove the filter -> The list would scroll away from the currently selected item
   f) Filter -> It was still possible to scroll down as if the list was not slimmed down

----

[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=44973&page=0#437861)


EvidentlyCube:
Keeping it as draft because it will update MIDs

BackEndLib/Wchar.cpp
```diff
@@ -885,3 +886,44 @@ WSTRING WCSReplace(
...
+	{
+		if (!token.empty())
+			result.push_back(std::move(token));
+	}
+
+	return result;
+}
+
+//*****************************************************************************
+bool WCSContainsAll(WSTRING const &haystack, std::vector<WSTRING> const needle)
```
schikore, line 919: Probably want needle to be a reference.

FrontEndLib/ListBoxWidget.cpp
```diff
@@ -948,6 +958,19 @@ void CListBoxWidget::Paint(
 		yDraw += CY_LBOX_ITEM;
 	}
 
+	if (this->wstrActiveFilter.size() > 0) {
+		const SURFACECOLOR FilterSeparatorColor = GetSurfaceColor(GetDestSurface(), 102, 102, 102);
+		
+		const UINT wTextY = this->y + DY_TEXT + this->h - CY_LBOX_ITEM - 1;
+		WSTRING message = L"Filter: ";
```
schikore, line 965: Extract to a .uni file

FrontEndLib/ListBoxWidget.cpp
```diff
@@ -880,29 +890,34 @@ void CListBoxWidget::Paint(
...
+#endif
+
 	//Draw inset area where text appears.
 	SDL_Surface *pDestSurface = GetDestSurface();
 	DrawInset(this->x, this->y, this->w, this->h, this->images[0],
 			pDestSurface, true, false, !IsEnabled());
 
-	if (this->Items.empty())
-		return;
+	const vector<LBOX_ITEM *> items = this->filteredItems;
```
schikore, line 904: Can this also be a reference so we don't copy the vector?  As it's const, I'd think so?

FrontEndLib/ListBoxWidget.cpp
```diff
@@ -1090,7 +1116,32 @@ void CListBoxWidget::HandleKeyDown(
...
+		case SDLK_BACKSPACE:
+			if (this->wstrActiveFilter.size() > 0) {
+				UpdateFilter(this->wstrActiveFilter.substr(0, this->wstrActiveFilter.size() - 1));
+				RequestPaint();
+			}
+			break;
+
+		case SDLK_ESCAPE:
+			if (this->wstrActiveFilter.size() > 0) {
+				UpdateFilter(L"");
```
mrimer, line 1128: In place of L"", use wszEmpty wherever this situation comes up

FrontEndLib/ListBoxWidget.cpp
```diff
@@ -1090,7 +1116,32 @@ void CListBoxWidget::HandleKeyDown(
...
+				RequestPaint();
+				PreventBubbling();
+			}
+			break;
+
 		default:
+			if (this->bAllowFiltering && !(KeyboardEvent.keysym.mod & KMOD_SHIFT)) {
+				const WCHAR character = KeyboardEvent.keysym.sym == SDLK_SPACE ? L' ' : TranslateUnicodeKeysym(KeyboardEvent.keysym);
```
mrimer, line 1136: I'm not sure that L' ' works outside of Windows on Linux and OS X.  We should probably use We(' ') instead for portability.

FrontEndLib/Widget.h
```diff
@@ -276,6 +276,7 @@ class CWidget : public CAttachableObject
 	virtual bool         Load();
 	bool        LoadChildren();
 	void        PaintChildren(bool bUpdateRects = false);
+	void        PreventBubbling();
```
mrimer, line 279: Suggest rename to clarify context: PreventEventBubbling or PreventEventBubblingUp

Also, suggest adding a parameter `bool bFlag = true`, so this call may also be made to re-enable bubbling by providing a `false` argument.

FrontEndLib/Widget.h
```diff
@@ -442,6 +443,8 @@ class CWidget : public CAttachableObject
 private:
 	void        DrawRectClipped(const SDL_Rect &rect, const SURFACECOLOR &Color,
 			SDL_Surface *pDestSurface);
+
+	bool bIsPreventingBubbling; // Set to true in event handlers to prevent current event being called on parent
```
mrimer, line 447: Suggest rename: bPreventEventBubbling.
I'd recommend moving this comment to the Prevent(Event)Bubbling method (to clarify that's actually where the call is made to set this to true) and have something like the following comment here:
`// when true, don't pass received events on to parent object`

mrimer (changes requested):
I love this functionality!
Expect we'll want to enable it in these same places for RPG also.

FrontEndLib/ListBoxWidget.cpp
EvidentlyCube: I have fixed this for all the instances of L"" and L'' that I've introduced in 5.2 so far.

EvidentlyCube:
@mrimer Mids updated, and all things fixed. I also found a few texts that had to be moved to MIDs from keymap functionality. Undrafting since it's now ready to review and merge

mrimer (approved):

PR #354, opened 2020-11-24
RPG: Add script expression parsing support for primitive functions with parameter lists.

mrimer:
Implement _EnemyStat(x,y,stat) primitive for use in script expressions.

Super easy to add new primitives going forward  See [here](https://github.com/CaravelGames/drod/pull/354/files#diff-7236b719adebbea66df2152d26dee8766a53c30ae1d75c7f657041cf0f047f30R140).

Edit: more primitives added

### DIFF
--- a/drodrpg/DRODLib/Character.cpp
+++ b/drodrpg/DRODLib/Character.cpp
@@ -42,6 +42,7 @@
 #include <BackEndLib/Ports.h>
 #include <BackEndLib/Files.h>
 
+//*****************************************************************************
 #define NO_LABEL (-1)
 
 #define NO_OVERRIDE (UINT(-9999))
@@ -123,9 +124,19 @@ const UINT MAX_ANSWERS = 9;
 
 #define SKIP_WHITESPACE(str, index) while (iswspace(str[index])) ++index
 
+//*****************************************************************************
 inline bool isVarCharValid(WCHAR wc)
 {
 	return iswalnum(wc) || wc == W_t('_');
+}
+
+void LogParseError(const WCHAR* pwStr, const char* message)
+{
+	CFiles f;
+	string str = UnicodeToUTF8(pwStr);
+	str += ": ";
+	str += message;
+	f.AppendErrorLog(str.c_str());
 }
 
 //*****************************************************************************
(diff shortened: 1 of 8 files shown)


mrimer:
@EvidentlyCube , do you think function primitives would be nice to have in 5.2 in script expressions?

Hypexion:
While I'm still not completely convinced about this specific case from a user experience perspective, I have to admit it would be really elegant and nice once you actually understand it. I'm in favour of putting this out there when the 1.3 alpha starts, and seeing how people react. If there's widespread confusion, it might be worth making a script command that does this, but if people generally get it I don't think that would be required.

I'd really like to see this in 5.2, since getting a few functions in would make scripting certain things a lot easier. Primate functions that I think would be useful in both games are `_Absolute()`, `_Minimum()` and `_Maximum()`, for doing some quick comparisons without need several lines of script.

EvidentlyCube:
I'd love to have it in classic DROD!

Just a note, I would be in favor of short function names mainly because of how narrow the text input is (which will be changing in 5.2 for sure).

Things I'd like to see:
 - `_abs(x)`
 - `_min(x,y)`
 - `_max(x,y)`
 - `_rotateCW(orientation)` - should just return the same number if invalid orientation
 - `_rotateCCW(orientation)` - should just return the same number if invalid orientation
 - Possibly all three distance calculations, likely as `_dist0()`, `_dist1()` and `_dist2()` for Chebyshev (L-inf), Manhattan (L-1) and Euclidean (L-2)

mrimer:
> Just a note, I would be in favor of short function names mainly because of how narrow the text input is (which will be changing in 5.2 for sure).

I like all the function names you provide here.  In the new Set monster var command, I provide a nice new, shiny, big text box for entering expressions. I also added a (new) label "Value or expression" to the new text box, to describe what can be entered there. 
 I suggest we apply similar labelling in the context of "Set var/wait until var", etc., in 5.2 and RPG for clarity.

I agree it would be nice to change the size of the tiny operand value box for variable operands.  Check out how the new "Set monster var" fields are displayed for possible inspiration (which is likely what you already have in mind anyway).

mrimer:
I've added all these primitives, plus three more:
```
_orient(dx,dy) //_o is already taken
_ox(o)
_oy(o)
```
Take a look at the second commit in the PR to see how easy it is to add a bunch of these.

mrimer:
Just realized there's a bug with referencing the player's position an orientation vars _X, _Y, and _O.  Will fix this on the main branch.

PR #357, opened 2020-11-25
Command list improvs

EvidentlyCube:
This depends on changes from #353.
[Relevant thread](http://forum.caravelgames.com/viewtopic.php?TopicID=39965&page=0#437953)

### DIFF
--- a/DROD/CharacterDialogWidget.cpp
+++ b/DROD/CharacterDialogWidget.cpp
@@ -26,6 +26,7 @@
  * ***** END LICENSE BLOCK ***** */
 
 #include "CharacterDialogWidget.h"
+#include "CommandListBoxWidget.h"
 #include "DrodBitmapManager.h"
 #include "DrodFontManager.h"
 #include "EditRoomScreen.h"
@@ -189,6 +190,9 @@ const UINT LIST_LINE_HEIGHT = 22;
 const SURFACECOLOR PaleRed = {255, 192, 192};
 
 std::map<UINT, UINT> CCharacterDialogWidget::speechLengthCache;
+const UINT CCharacterDialogWidget::INDENT_PREFIX_SIZE = 8;
+const UINT CCharacterDialogWidget::INDENT_TAB_SIZE = 3;
+const UINT CCharacterDialogWidget::INDENT_IF_CONDITION_SIZE = 5;
 
 #define NOT_FOUND (UINT(-1))
 
@@ -493,8 +497,8 @@ CCharacterDialogWidget::CCharacterDialogWidget(
 
 	AddWidget(new CLabelWidget(0L, X_COMMANDSLABEL, Y_COMMANDSLABEL,
 			CX_COMMANDSLABEL, CY_COMMANDSLABEL, F_Small, g_pTheDB->GetMessageText(MID_Commands)));
-	this->pCommandsListBox = new CListBoxWidget(TAG_COMMANDSLISTBOX, X_COMMANDS, Y_COMMANDS,
-			CX_COMMANDS, CY_COMMANDS, false, true, true);
+	this->pCommandsListBox = new CCommandListBoxWidget(TAG_COMMANDSLISTBOX, X_COMMANDS, Y_COMMANDS,
+			CX_COMMANDS, CY_COMMANDS);
 	AddWidget(this->pCommandsListBox);
 
 	//Appearance (character/tile graphic).
@@ -2462,7 +2466,7 @@ void CCharacterDialogWidget::OnKeyDown(
 						line!=selectedLines.end(); ++line)
 				{
 					if (*line < pCommands->size())
-						wstrCommandsText += toText(*pCommands, (*pCommands)[*line]);
+						wstrCommandsText += toText(*pCommands, (*pCommands)[*line], *line);
 				}
 				if (!wstrCommandsText.empty())
 					g_pTheSound->PlaySoundEffect(SEID_POTION);
@@ -3227,8 +3231,6 @@ WSTRING CCharacterDialogWidget::GetCommandDesc(
 const
 {
 	WSTRING wstr;
-
-	wstr += GetPrettyPrinting(commands, pCommand, 6, 3); //indent the If conditions considerably
 
 	//Call language-specific version of method.
 	switch (Language::GetLanguage())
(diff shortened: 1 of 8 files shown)


mrimer:
I've merged #353. Would you please rebase?

FrontEndLib/ListBoxWidget.cpp
```diff
@@ -1346,6 +1401,9 @@ void CListBoxWidget::HandleDrag(
 	} else {
 		this->wCursorLine = this->wDraggingLineNo;
 	}
+
+	UpdateFilter(L"");
```
mrimer, line 1405: L"" --> wszEmpty

BackEndLib/Wchar.cpp
```diff
@@ -130,6 +131,13 @@ void AsciiToUnicode(const std::string &str, WSTRING &wstr) {
 	AsciiToUnicode(str.c_str(), wstr);
 }
 
+//*****************************************************************************
+const WSTRING AsciiToUnicode(const char *psz) {
```
SuperV1234, line 135: Returning by `const` value doesn't really help here, assuming `WSTRING` is the same as [this one](https://github.com/CaravelGames/drod/blob/master/BackEndLib/Wchar.h#L77).

```suggestion
WSTRING AsciiToUnicode(const char *psz) {
```

I would also look into `[[nodiscard]]` for function signatures like these. It is a C++17 feature but there are equivalent compiler-specific attributes since C++98. It produces a warning when the return value of `AsciiToUnicode` is discarded, avoidining mistakenly assuming that `psz` gets mutated rather than a new value being returned:

```suggestion
[[nodiscard]] WSTRING AsciiToUnicode(const char *psz) {
```

BackEndLib/Wchar.cpp
```diff
@@ -885,3 +893,45 @@ WSTRING WCSReplace(
...
+//*****************************************************************************
+WSTRING WCSToLower(WSTRING const &source)
+//Converts WSTRING to lowercase
+{
+	WSTRING lowercased;
+	lowercased.reserve(source.size());
+	for (WSTRING::const_iterator it = source.begin(); it != source.end(); ++it)
+		lowercased += towlower(*it);
+
+	return lowercased;
```
SuperV1234, line 906: C++11 range-based `for` is cleaner here:

```suggestion
	for (char c : source) 
		lowercased += towlower(c);

	return lowercased;
```

BackEndLib/Wchar.cpp
```diff
@@ -885,3 +893,45 @@ WSTRING WCSReplace(
...
+}
+
+//*****************************************************************************
+const std::vector<WSTRING> WCSExplode(WSTRING const &source, WCHAR const delimiter)
```
SuperV1234, line 910: Ditto, returning by `const` value is not useful and inhibits move semantics. See https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#f20-for-out-output-values-prefer-return-values-to-output-parameters

BackEndLib/Wchar.cpp
SuperV1234: The `[[nodiscard]]` suggestion applies to pretty much every function here, or every function returning a value that should never be ignored.

BackEndLib/Wchar.cpp
```diff
@@ -885,3 +893,45 @@ WSTRING WCSReplace(
...
+bool WCSContainsAll(WSTRING const &haystack, std::vector<WSTRING> const &needles)
+// Returns true if haystack contains every string in needle, even if they overlap
+{
+	for (std::vector<WSTRING>::const_iterator iter = needles.begin(); iter < needles.end(); ++iter)
+	{
+		if (haystack.find(*iter) == WSTRING::npos)
+			return false;
+	}
+
+	return true;
```
SuperV1234, line 936: From `<algorithm>`, C++11:

```cpp
	return std::all_of(needles.begin(), needles.end(),
		[&haystack]
		{
			return haystack.find(*iter) != WSTRING::npos;
		});
```

Consider `std::none_of` as well, might be more readable due to the avoided negation in the body of the lambda expression.

DROD/CharacterDialogWidget.cpp
```diff
@@ -189,6 +190,9 @@ const UINT LIST_LINE_HEIGHT = 22;
 const SURFACECOLOR PaleRed = {255, 192, 192};
 
 std::map<UINT, UINT> CCharacterDialogWidget::speechLengthCache;
+const UINT CCharacterDialogWidget::INDENT_PREFIX_SIZE = 6;
+const UINT CCharacterDialogWidget::INDENT_TAB_SIZE = 3;
+const UINT CCharacterDialogWidget::INDENT_IF_CONDITION_SIZE = 5;
```
SuperV1234, line 195: If these are purely hardcoded constants, consider using a [`constexpr` variable](https://en.cppreference.com/w/cpp/language/constexpr) (C++11), or an `inline constexpr` variable if C++17 support is available. Just makes it more obvious and has a stronger guarantee that the value is purely known at compile-time.

SuperV1234 (commented):
Nice to see DROD being developed again. Some pedantic C++ stuff to improve code quality/readability. I'm not sure what C++ standard is targeted, but pretty much every platform in the world supports C++11 nowadays, and most support C++14 and C++17 as well.

EvidentlyCube:
@SuperV1234  I am afraid we're currently not targeting C++11 :(.

SuperV1234:
> @SuperV1234 I am afraid we're currently not targeting C++11 :(.

What platform is stopping you? Seems a bit of a weird decision in 2020.

SuperV1234:
Also, https://github.com/CaravelGames/drod/blob/master/Master/Linux/ninjamaker#L150

mrimer:
> > @SuperV1234 I am afraid we're currently not targeting C++11 :(.
> 
> What platform is stopping you? Seems a bit of a weird decision in 2020.

Our current OS X development setup.

EvidentlyCube:
@mrimer  PR updated, I also fixed labels not having negative indent.

DROD/CharacterDialogWidget.cpp
```diff
@@ -189,6 +190,9 @@ const UINT LIST_LINE_HEIGHT = 22;
 const SURFACECOLOR PaleRed = {255, 192, 192};
 
 std::map<UINT, UINT> CCharacterDialogWidget::speechLengthCache;
+const UINT CCharacterDialogWidget::INDENT_PREFIX_SIZE = 8;
```
mrimer, line 193: These constants should be consolidated with the analogous magic numbers featured elsewhere in the code, e.g., CCommandListBoxWidget.

mrimer (changes requested):

Hypexion:
@mrimer In light of skell's departure from DROD development, could this be merged as-is? The feature is complete, and I can whip up a PR to move the magic number to live with its friends.

mrimer:
Merging. My last request can be applied as a follow-on PR.

DROD/CharacterDialogWidget.cpp
Hypexion: @mrimer I'm not quite sure what needs to be consolidated where. These constants appear to be grouped with the other ones in this file, and they don't appear to share a purpose with the ones in CCommandListBoxWidget. That file does have some constants littered about, but they are in the place where they're being used, so it doesn't seem like much of an issue.

mrimer: I appreciate you looking into this.  I possibly misinterpreted the values.  IIRC, to me, it looked like there are identical dimension values in both places, e.g., 8 pixels for an indent size, that possibly need to remain in lockstep in both places for things to work correctly.  If this isn't the case, I'm fine with leaving things as-is for now.

PR #371, opened 2020-12-10
RPG: Fix regression in CCharacter::IsTileObstacle()

Hypexion:
Due to changes related to setting character movement, wall and water dwelling characters can no longer move over regular floors, which is wrong. This PR fixes this.


mrimer (approved):

mrimer:
Thanks, @Hypexion !

PR #372, opened 2020-12-16
Set movement type script command

Hypexion:
`Set movement type` allows the movement of a character to be changed. This involves some changes to `CCharacter::IsTileObstacle()` to ensure backwards compatibility in addition to ensuring changing the movement type has an effect.

This also adds a new behavior, `Restricted Movement`. This causes characters with the Water and Wall movement types to be unable to move outside of their "mediums", reflecting how these movement types work for non-scripted monsters.


mrimer (approved):

mrimer:
Thank you, @Hypexion !
Is there a forum topic for this feature request?

Hypexion:
> 
> 
> Thank you, @Hypexion !
> Is there a forum topic for this feature request?

There isn't a thread, although this could be consider tangental to the behaviors thread.

PR #374, opened 2020-12-19
RPG: Characters as monsters don't have a name

Hypexion:
Thread: http://forum.caravelgames.com/viewtopic.php?TopicID=45005

The case where a character had an identity less than `CHARACTER_FIRST` was missed, meaning regular monster characters did not return the correct name.

### DIFF
--- a/drodrpg/DROD/RoomWidget.cpp
+++ b/drodrpg/DROD/RoomWidget.cpp
@@ -1803,9 +1803,8 @@ WSTRING CRoomWidget::GetMonsterName(CMonster* pMonster) const
 				//custom character type was deleted, and this NPC's type left dangling
 				wstr += wszQuestionMark;
 			}
-		} else {
+		} else if (pCharacter->wIdentity >= CHARACTER_FIRST) {
 			bCharacterName = true;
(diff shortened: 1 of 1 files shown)


mrimer:
Thank you, @Hypexion !

This change should be merged onto the main branch instead of 1.3, correct?

Hypexion:
> 
> 
> Thank you, @Hypexion !
> 
> This change should be merged onto the main branch instead of 1.3, correct?

I'm pretty sure this relies on other 1.3 changes, but I can double check to be certain.

mrimer:
> > Thank you, @Hypexion !
> > This change should be merged onto the main branch instead of 1.3, correct?
> 
> I'm pretty sure this relies on other 1.3 changes, but I can double check to be certain.

Ah, sorry for any confusion.  You may be right, but please double check if you would.

Hypexion:
@mrimer I tried cherry picking this change onto the main branch, and it lead to a conflict. Examining it gives me the impression that this fix is only needed after the addition of the _MyName functionallity.

mrimer (approved):

mrimer:
Thank you for confirming!  Approved and merged.

PR #380, opened 2021-03-24
Ally target behavior

Hypexion:
A behavior to make Stalwarts and Soldiers attack an NPC. Characters are now a type of monster Stalwarts can target, and `CMonster` now has the `IsPlayerAllyTarget()` function, which is overridden by `CCharacter`. For normal monsters, `IsPlayerAllyTarget()` simply wraps the current condition of `IsAlive()`. For NPCs, it will check for the Ally Target behavior.


DRODLib/Stalwart.cpp
```diff
@@ -217,7 +217,7 @@ bool CStalwart::GetGoal(UINT& wX, UINT& wY)
 	const CMonster *pMonster = this->pCurrentGame->pRoom->pFirstMonster;
 	while (pMonster)
 	{
-		if (CStalwart::typesToAttack.has(pMonster->GetIdentity()))
+		if (CStalwart::typesToAttack.has(pMonster->wType) && pMonster->IsPlayerAllyTarget())
```
mrimer, line 220: Pardon, don't have the project open in front of me.  Am I correct in understanding this change will not alter pre-existing logic of how stalwarts will target characters with a certain appearance?

That is, previously, a stalwart won't attack an NPC of type Roach, and now it still won't, unless the Ally Target flag is set, correct?

Data/Help/1/script.html
```diff
@@ -530,12 +530,13 @@ <h3><u>Character Script Commands</u></h3>
 	<li><a name="beh-drop-trapdoor"><b>Drops Trapdoors</b></a>: The NPC will drop trapdoors and melt thin ice.</li>
 	<li><a name="beh-drop-trapdoor-armed"><b>Drop Trapdoors When Armed</b></a>: The NPC will drop trapdoors if it is armed with a weapon, and will melt thin ice.</li>
 	<li><a name="beh-push-objects"><b>Push Objects</b></a>: The NPC can push objects such as Mirrors and Powder Kegs.</li>
-	<li><a name="beh-push-monsters"><b>Push Objects</b></a>: The NPC can push monsters that can be pushed. Currently, only NPCs with <a href="#imp-pushable-by-body">Pushable by body</a> can be pushed in this manner.</li>
+	<li><a name="beh-push-monsters"><b>Push Monsters</b></a>: The NPC can push monsters that can be pushed. Currently, only NPCs with <a href="#imp-pushable-by-body">Pushable by body</a> can be pushed in this manner.</li>
```
mrimer, line 533: Good catch! Thank you.

mrimer (approved):
This looks good to me.  Thank you!

I have one clarifying question to refresh my memory.

DRODLib/Stalwart.cpp
Hypexion: This is correct. NPCs can now be considered for targeting, but the behavior flag is required for them to actually "become" a target. Otherwise, Stalwarts and Soldiers remain as blind to NPCs as they've always been.

PR #381, opened 2021-04-17
Additional object selection highlighting

Hypexion:
Stalwart, Soldier and Citizen pathfinding is weird and confusing to many players. So there is a desire to visualize it. It would also be nice if clicking on horns showed where the spawn will occur.

Monster path visuals extend the implementation of #359, allowing paths to be viewed for Stalwarts, Soldiers, Citizens and Engineers.

### DIFF
--- a/DROD/RoomWidget.cpp
+++ b/DROD/RoomWidget.cpp
@@ -1041,10 +1041,11 @@ void CRoomWidget::HighlightSelectedTile()
 				UINT wDestX, wDestY;
 				if (pCitizen->GetGoal(wDestX, wDestY)) {
 					AddShadeEffect(wDestX, wDestY, PaleYellow);
-					for (UINT wPathIndex = 0; wPathIndex < pCitizen->GetStoredPath().GetSize(); ++wPathIndex)
+					CCoordStack citizenPath = pCitizen->GetStoredPath();
+					for (UINT wPathIndex = 0; wPathIndex < citizenPath.GetSize(); ++wPathIndex)
 					{
 						UINT wX, wY;
-						pCitizen->GetStoredPath().GetAt(wPathIndex, wX, wY);
+						citizenPath.GetAt(wPathIndex, wX, wY);
 						ASSERT(this->pRoom->IsValidColRow(wX, wY));
 						AddShadeEffect(wX, wY, PaleYellow);
 					}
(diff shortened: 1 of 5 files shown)


DROD/RoomWidget.cpp
```diff
@@ -1039,8 +1039,16 @@ void CRoomWidget::HighlightSelectedTile()
 				//Show citizen's current destination.
 				const CCitizen *pCitizen = DYN_CAST(const CCitizen*, const CMonster*, pMonster);
 				UINT wDestX, wDestY;
-				if (pCitizen->GetGoal(wDestX,wDestY))
+				if (pCitizen->GetGoal(wDestX, wDestY)) {
 					AddShadeEffect(wDestX, wDestY, PaleYellow);
+					for (UINT wPathIndex = 0; wPathIndex < pCitizen->GetStoredPath().GetSize(); ++wPathIndex)
```
mrimer, line 1044: I didn't mention this before on the Halph path PR, but the code calling `GetStoredPath()` would be more efficient and cleaner if the result were saved in a local variable, so the data structure isn't re-copied from the monster object twice in each iteration of these loops (an N-squared cost).

DROD/RoomWidget.cpp
```diff
@@ -1155,6 +1170,26 @@ void CRoomWidget::HighlightSelectedTile()
 			bRemoveHighlightNextTurn = false;
 		}
 		break;
+		case T_HORN_SQUAD:
+		{
+			MovementType eMovement = GetHornMovementType(this->pCurrentGame->swordsman.GetMovementType());
```
mrimer, line 1175: Though I haven't checked, it feels like there is some duplication of game engine logic here.

For improved maintainability, it may make sense to wrap this logic in a method call, possibly in CCurrentGame.  For instance, refactor the code getting goalX/Y for each of the two horns into a method in DRODLib named something like `GetNearestEntranceForHorn(wX,wY)`, and call that method here for each case.

mrimer (changes requested):

DROD/RoomWidget.cpp
Hypexion: I was able to trim out the movement type stuff into a general function for finding horn entrances for a given entity type. It makes `CCurrentGame::BlowHorn` a little tidier, and probably makes add more horns slightly easier.

mrimer (approved):

mrimer:
Thank you, @Hypexion !

PR #382, opened 2021-04-29
Add support for token build markers

Hypexion:
Allows token build markers to be plotted and to appear correctly. Also adds all token types other than the used temporal split to the list of buildable objects in the script editor.

Minor changes have been done in `RoomWidget` so that tooltips and effects for token build markers work.

See thread: http://forum.caravelgames.com/viewtopic.php?TopicID=45196


mrimer (commented):
Thank you for contributing this improvement :)

One comment and question: if I'm understanding correctly, this new functionality will not render in the room in present-day 5.1 builds, i.e., if using an older 5.1 engine to play a hold created by a user applying this new functionality into its design.   If that's so, maybe defer this addition till 5.2?

Hypexion:
> One comment and question: if I'm understanding correctly, this new functionality will not render in the room in present-day 5.1 builds, i.e., if using an older 5.1 engine to play a hold created by a user applying this new functionality into its design. If that's so, maybe defer this addition till 5.2?

Ah, I didn't of that. Without the logic change, the build markers won't function. I'll get a new PR ready for this that's for 5.2

PR #386, opened 2021-05-11
Fix improved script editor to work with default script editing

Hypexion:
#357 adds a number of visual improvements to script editing. However, it was overlooked that `CCharacterDialogWidget` actually has _two_ lists of commands - one for the character's script, and one for editing default scripts. The changes for #357 only apply to the main editor's listbox, meaning that it will be incorrectly referenced while editing default scripts. This crashes DROD due to out-of-bounds errors if the default script being edited is longer than the character script.

I've gone through and fixed that, and also updated the default script editor's list to be a `CCommandListBoxWidget`


mrimer (approved):
Thank you!

PR #387, opened 2021-05-11
"Replace with Default Script" command

Hypexion:
The oft desired command to clear out a character's command buffer and replace it with it's default script. Doesn't do anything for non-custom NPC types.

Related forum post: http://forum.caravelgames.com/viewtopic.php?TopicID=41035&page=0#405497


mrimer (approved):
Thank you!

PR #388, opened 2021-05-29
RPG: Fix crash when clicking bomb in editor

Hypexion:
Preview bomb explosion calls CDbRoom::ConvertUnstableTar(). This function needs an active game to work, so a crash happens. Confusingly, it only crashs in the alpha. Anyhow, I updated the function to match the mainline version for getting player co-ordinates to avoid a crash.

Thread: http://forum.caravelgames.com/viewtopic.php?TopicID=45237


mrimer (approved):
Thank you for the fix!

PR #389, opened 2021-07-13
"Set Var At" script command

Hypexion:
`Set Var At` is a script command that allows an character to invoke `CCharacter::SetVariable()` on another NPC at a given room position. This results in variables being set as if the target NPC had executed a `Set Var` command. This means any local character variables will be written or read for the target NPC, allowing for an increased access to these variables.

Related thread:  https://forum.caravelgames.com/viewtopic.php?TopicID=41035


Data/Help/1/script.html
```diff
@@ -266,6 +266,8 @@ <h3><u>Character Script Commands</u></h3>
       from the hold and all NPC scripts that use it, not just from this
       particular script.<br />
 </li>
+  <li><a name="setvarat"><b>Set var at</b></a> - Allows the <a href="#setvar">Set var</a> command to be invoked with an NPC on another tile. All assignment and parsing will occur as if the target NPC called the Set var command. Thus, it's local variables will be used, allowing them to be written or read remotely. This command can also be used an <a href="#if">If</a> conidtion. The condition is statified if the command successfully invokes a variable assingment on an NPC in the target square. Hint: Use this for scripts that rely on the result of a remote variable assignment.
```
mrimer, line 269: it's local -> its local
conidtion -> condition
statified -> satisfied
assingment -> assignment

DRODLib/Character.cpp
```diff
@@ -2990,6 +2992,41 @@ void CCharacter::Process(
...
+				//Remotely set local variable w (with operation h) of NPC at (x,y)
+				UINT wX, wY;
+				bool success = false;
+				getCommandXY(command, wX, wY);
+
+				CMonster* pMonster = room.GetMonsterAtSquare(wX, wY);
+				if (pMonster) {
+					CCharacter* pCharacter = DYN_CAST(CCharacter*, CMonster* ,pMonster);
+					if (pCharacter) {
+						// Transfer var set information to a tempory command.
```
mrimer, line 3006: tempory -> temporary

DRODLib/CharacterCommand.h
```diff
@@ -324,6 +324,7 @@ class CCharacterCommand
 		CC_PushTile,            //Push object or entity at (x,y) in direction (w)
 		CC_SetMovementType,     //Sets the character's movement type to X.
 		CC_ReplaceWithDefault,  //Replaces the character's script with its default script (if possible)
+		CC_VarSetAt,            //Remotely set local variable w (with operation h) of NPC at (x,y)
```
mrimer, line 327: Let's add one more detail here, including how flags is used as a relevant parameter, e.g.,
`//Remotely set local variable w (with operation h) of NPC at (x,y) to value in flags`

mrimer (commented):
Thank you for implementing this new script command!
I can see this feature making script execution more complicated, but it could also make room script management simpler.

I just noted some spelling issues to correct before merging.

mrimer (approved):

PR #390, opened 2021-07-13
Don't increment command index after default script replacement

Hypexion:
Fix for minor problem in #387, where the first command in a default script will be skipped due to the command index incrementing after the script is replaced with the default script. Automatic increment is supressed for the command. If the replacement fails, the index is manually incremental to prevent the script runner getting stuck.


DRODLib/Character.cpp
```diff
@@ -2896,6 +2896,9 @@ void CCharacter::Process(
 					wTurnCount = 0;
 					++wVarSets; //Count as setting a variable for loop avoidence
 					LoadCommands(this->pCustomChar->ExtraVars, this->commands);
+				}	else {
```
mrimer, line 2899: Replace internal tab with space

mrimer (approved):

PR #391, opened 2021-07-17
RPG: "Replace with Default Script" & "Set var at" commands

Hypexion:
Implementation of #387 and #389 in DROD RPG.

`Replace with Default Script` removes the content of NPC's script buffer and replace's it with the NPC's default script if possible.

`Set var at` invokes `CCharacter::SetVariable()` on another NPC at a given room position. This allows access to NPC variables not covered by `Set monster var` and the `_EnemyStat()` primitive function.

Related thread: http://forum.caravelgames.com/viewtopic.php?TopicID=32862


mrimer (commented):
Thank you!
Really appreciate you migrating the new commands you added to RPG as well.
Just making a note here to apply the fixes I noted in the other PRs here also.

One other detail: IIRC, local variables are not implemented in RPG as they are in 5.x.  Assuming you're aware of that, I could see how this command would still be useful to modify remote preset monster variables.  Is that the intent?  If so, then I think the `SetMonsterVar` command should already handle this use case.

Your thoughts?

Hypexion:
> 
> One other detail: IIRC, local variables are not implemented in RPG as they are in 5.x. Assuming you're aware of that, I could see how this command would still be useful to modify remote preset monster variables. Is that the intent? If so, then I think the `SetMonsterVar` command should already handle this use case.
> 
> Your thoughts?

`SetVarAt` also allows a scripter to access a character's various `_MyFoo` variables in addition to it's statistics. I think this provides some utility for people who want to do more estoric things with the engine, as well as the goofy "item that pushes NPCs around" option. It's not that much, but I can imagine that someone might find it useful.

mrimer:
> `SetVarAt` also allows a scripter to access a character's various `_MyFoo` variables in addition to it's statistics. I think this provides some utility for people who want to do more estoric things with the engine, as well as the goofy "item that pushes NPCs around" option. It's not that much, but I can imagine that someone might find it useful.

Ah, okay.  I couldn't recall whether SetVarAt covered all the _MyFoo vars.  This sounds good.

I'll approve the PR when those minor fixes get committed.

mrimer (approved):

PR #392, opened 2021-07-18
Support moods for custom portraits

Hypexion:
A small change to face drawing code to support moods for custom portraits. The face's mood value is used to offset the source rectangle (provided the image is wide enough), allowing a custom portrait to provide graphics for moods other than normal.  In additional to speaking moods, it also means that a custom dying mood can be provided for custom player roles.

Related thread: http://forum.caravelgames.com/viewtopic.php?TopicID=41611


mrimer (approved):
Wow, thanks!
What an elegant solution.

PR #393, opened 2021-07-18
RPG: Support moods for custom portraits

Hypexion:
Updates the face management code in DROD RPG to the newer DROD 5.0 face management. This makes faces better, I assume.

Also includes the small change to face drawing code to support moods for custom portraits. The face's mood value is used to offset the source rectangle (provided the image is wide enough), allowing a custom portrait to provide graphics for moods other than normal.


mrimer (approved):
Yes, it is helpful to remove the technical debt in the old face code remaining in the RPG codebase.  Thank you!

Issue #71, opened 2020-09-29
TODO Before 5.2


EvidentlyCube:
Similarly to #70 this is a topic where I'll list things and new features I'd like to have implemented as part of 5.2. For things which have no thread I'll make one soonish,

The reason I want those for 5.2 is 100% selfish - most of the things here would improve the experience of the hold I am working on.

- [ ] Build and Remove checkpoints.
- [ ] [Unlock editing rights script command](http://forum.caravelgames.com/viewtopic.php?TopicID=40423)
- [ ] Play custom sound effect command (as opposed to using Speech which can only play one thing at a time)
- [ ] Room Position in level variables
- [ ] Custom names for characters
- [ ] [Allow ghost display objects to be right-clicked](http://forumskell.caravelgames.com/viewtopic.php?TopicID=44328)
- [ ] Replace image instead of Remove -> Return to level screen -> add again -> reconnect the image everywhere it was used
- [ ] [Emotive Character Portraits](http://forumskell.caravelgames.com/viewtopic.php?TopicID=41611)
- [ ] A bunch of image overlay improvements:
   - `Group <number|default=0>` which assigns a group ID to an image
   - `ClearGroup <number>` similar to `ClearLayer` removes images in the specified group
   - `DisplayRectMod <deltaX> <deltaY> <deltaW> <deltaH>` similar to `DisplayRect` but instead of setting the values it modifies them
   - `AddX <deltaX>` and `AddY <deltaY>` modifies the position by delta immediately (alternatively make `Move` with 0 time do the same)

Hypexion:
The words "remove checkpoints" fill me with uncontainable dread. While most DROD architects are good and reasonable people, I don't think this is the sort of power that should be given out.

EvidentlyCube:
> I don't think this is the sort of power that should be given out.
Is there anything this command could do that'd be worse than not having checkpoints in the first place? And it has a very legitimate use - if your hold has some kind of customized movement which requires teleporting the player a lot, which normally causes checkpoints to auto trigger every other turn even if you don't move... This could be avoided with this.

I'd be happy to also just have a "Trigger checkpoint now" command.

mrimer:
For "Play custom sound effect command", note that when a speech duration of 1ms is applied, you'll essentially get multiple sound effects playing on top of one another.

mrimer:
@Hypexion , I think you added this support, correct?

[...]

PR #396, opened 2021-07-20
Additional Image Overlay commands

Hypexion:
Implements most of the Image Overlay commands requested in #71:

`AddX` and `AddY` move the the image by the given amount in the given axis.
`DisplayRectModify` allows the source region of the image to be altered
`Group` and `CancelGroup` allows images to be assigned to a group that can be cleared at a later point.


mrimer (approved):
I appreciate you adding these new image commands!

PR #397, opened 2021-07-21
Fix clone switch command assignment

Hypexion:
This fixes a typo in the key to command mapping that means the command to switch the active player instead restarts the room.


mrimer (approved):

mrimer:
Thank you!

PR #398, opened 2021-07-26
Addition options for Attack Tile command

Hypexion:
Adds a few requested options for `Attack Tile`:

`AT_Remove` allows a monster to be removed from the room without trigger death effects. Useful for getting rid of critical monsters, or monsters that leave remains.
`AT_OneTurnStun` and `AT_TwoTurnStun` apply stun effects of the indicated lengths.
`AT_Floorspikes` and `AT_Firetrap` emulate the stabs of the two stabby room elements.

Related thread: http://forum.caravelgames.com/viewtopic.php?TopicID=44412


mrimer (approved):
Excellent!  Thank you!

PR #399, opened 2021-07-29
Split Swirl effect into player and tile swirl

Hypexion:
This adds a new Swirl effect with the swirl is centered on a tile rather than the player. The old effect is now named "Player swirl" in the scripting UI.

To achieve this, `CSwordsmanSwirlEffect` has been refactored to move most the actual swirl drawing to the new `CSwirlEffect`, and a new class `CTileSwirlEffect` has been added to support the tile-targeted swirl.

Related thread: http://forum.caravelgames.com/viewtopic.php?TopicID=40797


mrimer (approved):

mrimer:
Neat!  Thank you!

PR #401, opened 2021-08-03
"Smart" monsters treat puffs as pathing obstacles

Hypexion:
Citizens, Engineers, Halphs, Stalwarts, Soldiers and Slayers will now treat Puffs as obstacles, and pathfind around them. Relay Stations will also consider Puffs as obstacles when directing Citizen movement, but not when constructing pathmaps.

Some unit tests have been added to validate basic expectations of this behavior change.

Reporting thread: http://forum.caravelgames.com/viewtopic.php?TopicID=45266


mrimer (approved):
Thank you!

PR #408, opened 2021-08-13
Disable surface blending for tile images

Hypexion:
Disable surface blending on images that are used for room tiles.

See thread: http://forum.caravelgames.com/viewtopic.php?TopicID=45004


mrimer (approved):

mrimer:
That  was easy!  Thank you!

PR #416, opened 2021-09-13
Wait for open tile script command

Hypexion:
A script command to check if a tile at a specified position is open, for a specified movement type. Also allows weapons and various combinations of entities to be ignored, if required. This includes allowing Puffs to be ignored separately from other monsters. The option to ignore "Beethro" isn't included as that would make things generally messier and I don't think it has that much appplication.

This command also doesn't take arrow direction or orthosquares into account, but other 5.2 features make checking that remotely less of a hassle. If we aren't worried about Too Many Script Commands it's not to difficult to have another command to check for them.


mrimer (approved):
Looks great!

Issue #80, opened 2020-09-30
UNI file encoding


Hypexion:
Currently, the .uni files that contain all the game strings use UCS-2 encoding. It appears that very few tools recognise these files as being text files, including the Github web interface. This makes them difficult to work with, especially if a merge conflict occurs.

Once DROD 5.2 and RPG 1.3 are done, we might want to look into changing the encoding to one that's more widely supported.

EvidentlyCube:
I'd suggest UTF-8 which is de-facto standard everywhere nowadays. No clue how much of a pain it would be to use it in C++ though.

mrimer:
I agree this is a pain.  I believe this format is tightly coupled with the CDbMessageText database format that, IIRC, expects UCS-2.
I could be wrong about this, but that would likely be the place to look first to make a format change.

EvidentlyCube:
I think it's fine to keep the format internally, nothing stops us from converting from whatever format we use to UCS-2.

Hypexion:
Having done some very brief research on this, it seems like if the format is changed, we'd want to change to UTF-16. It seems to be the case that UCS-2 is an older implementation of the Unicode specification, that is obselete now that UTF-16 exists.

Poking around in the DROD source, all the string functions are using `WSTRING` and `WCHAR`. `Wchar.h` in BackEndLib claims that these are 16 bit unicode characters, so it should probably be fine to switch over to UTF-16 encoding.

The next step forward is probably to just do it, and see if the strings still get imported correctly by DRODUtil.

mrimer:
Sure!  What you say makes sense to me.  I don't want to stand in the way of progress.  Please go ahead and try it out if you like and if it works, then great!

mrimer:
> I think it's fine to keep the format internally, nothing stops us from converting from whatever format we use to UCS-2.

That's a great point too.  I often forget about interface boundaries.  We do have a bunch of conversion routines already, and we can do more of the same in WChar/DRODUtil if we wish without changing the internal DB format.

Hypexion:
Belated update: Having looked into this some more, the big issue is that git essentially doesn't support UTF-16 or UCS-2. Switching between them won't help. If we want the uni files to play nicely with Git, we'd probably want to convert them to UTF-8, then change DRODUtil to deal with that change.

mrimer:
Merged PR #428 into 5.2.  If we want, we may also do this for RPG.

Hypexion:
[...]

PR #428, opened 2022-01-08
Convert Uni files to UTF-8

Hypexion:
## What
Converts the `.uni` files containing game strings from UCS-2 to UTF-8.
Makes a change in DRODUtil so that it can correct read in the files in the new encoding.

## Why
Due to technical reasons related to how character encoding works, Git and Github don't recognise UCS-2 encoded files as text. This means they are treated as binary files, which do not benefit from various Git functions such as file comparison and merge conflict resolution. This makes developing a litte more painful when new strings are added, or if strings are changed (such as when a new translation is added). See #80 for more discussion.

## What in detail
Thanks to the ~~horrors~~ magic of C++, only a few small changes are required. Once a UTF-8 file is read in, the resulting `char` array can be converted to `WSTRING` using the function `UTF8ToUnicode()`. From there, the reading function can operate on the underlying wide character array accessed via `c_str()`.

## Testing
I've ran the updated DRODUtil on the converted `.uni` files, and it seems to be working for English. However, I don't know how to change DROD's language, and even if I did, I don't speak French or Russian, so I wouldn't know if strings were being mangled in a non-obvious way. The files themselves look correct, and I did have a peek at the results of `UTF8ToUnicode()` in the debugger, which looked fine. However, it might be worth someone else having a quick check for the other two languages.


mrimer (approved):
Thank you for the conversion!  This will make maintaining the .uni files so much nicer.  Glad this is all it took for this change to be implemented.  Kudos!

PR #430, opened 2022-01-08
RPG: Disable surface blending for tile images

Hypexion:
It's #408 but for DROD RPG. Arguably more important since the transparency colour in RPG is eye-searing magenta.

Related thread: http://forum.caravelgames.com/viewtopic.php?TopicID=45514


mrimer (approved):
Thank you!

PR #433, opened 2022-01-30
Logical Wait commands/Complex conditionals

Hypexion:
## Overview
Implements a set of script commands that allow for multiple conditional script commands to be associated into single logic blocks. These "Logical Wait" commands allow script execution to be halted until specific conditions are met, and allows conditional logic to be more easily compressed. Additionally, these logic blocks can be nested, for when complex logic is required.

Related thread: http://forum.caravelgames.com/viewtopic.php?TopicID=45498&page=0#442503

## Details
A logic block consists of one or more conditional commands, usually `Wait for` commands, enclosed between one of three `Logical Wait` commands and a `Logical Wait End` command. The possible operators are `And`, `Or` and `Exclusive Or` (other operators can be implemented later if desired). When the script processor encounters a logical wait command, it will evaluate the commands contained within the logic block to determine if execution should continue. If successful, the script will jump to the next command after the `Logical Wait End` command. If not, the script will halt until the next game turn.

Due to how script execution works, commands within a logic block are not processed normally. Instead, each logical wait command has its own function in `CCharacter`, and evaluation of conditional commands is performed by the new method `CCharacter::EvaluateConditionalCommand`. To enable this, the logic for supported conditional commands has been refactored into methods to allow for reuse. This means that commands that aren't supported in logic blocks do not run (unless the scripter puts a label inside a logic block).

(A bug related to the `Face towards like guard` command is also fixed in this PR. See thread: http://forum.caravelgames.com/viewtopic.php?TopicID=45536)

## UI
Much like the `If` blocks, logic blocks are indented in the script editor. Commands that are not supported in logic blocks are marked to alert the user to the issue.


DROD/CharacterDialogWidget.cpp
```diff
@@ -4577,6 +4596,11 @@ void CCharacterDialogWidget::PopulateCommandListBox()
 	this->pActionListBox->AddItem(CCharacterCommand::CC_WaitForRemains, g_pTheDB->GetMessageText(MID_WaitForRemains));
 	this->pActionListBox->AddItem(CCharacterCommand::CC_WaitForWeapon, g_pTheDB->GetMessageText(MID_WaitForWeapon));
 
+	this->pActionListBox->AddItem(CCharacterCommand::CC_LogicalWaitAnd, L"Logic Wait And");
```
mrimer: I'm brainstorming whether alternative command names might be more descriptive to the lay user.  For instance:

"Wait for All"
"Wait for Any"
"Wait for Exactly One"
"Wait for Conditions End"

or

"Wait for All of"
"Wait for Any of"
"Wait for Exactly One of"
"Wait for Expressions End"

etc.

Just a thought.

DRODLib/Character.cpp
```diff
@@ -4007,6 +3686,229 @@ bool CCharacter::IsPushableByWeaponAttack() const
...
+			player.wY >= py && player.wY <= py + ph)
+			return true;
+	}
+	if ((pflags & ScriptFlag::HALPH) != 0)
+	{
+		if ((player.wAppearance == M_HALPH || player.wAppearance == M_HALPH2) &&
+			player.wX >= px && player.wX <= px + pw &&
+			player.wY >= py && player.wY <= py + ph)
+			return true;
+		else if (room.IsMonsterInRectOfType(px, py,
```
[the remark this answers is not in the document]
Hypexion: That makes sense. Keeping around superflous code, even if it's just a few elses is something worth avoiding.

DROD/CharacterDialogWidget.cpp
Hypexion: Yeah, I'm thinking the UX on these names will probably be better, especially for users who don't have familarity with formal logic terms. The "of" versions provide more clarity in text, but might add a little extra clutter visual when editing as the script editor will be do some stuff to make these commands noticable.

Hypexion: Yeah, I'm thinking the UX on these names will probably be better, especially for users who don't have familarity with formal logic terms. The "of" versions provide more clarity in text, but might add a little extra clutter visual when editing as the script editor will be do some stuff to make these commands noticable.

Data/Help/1/script.html
```diff
@@ -361,7 +361,7 @@ <h3><u>Character Script Commands</u></h3>
...
+  <li><a name="logical-wait"><b>Logical Waits</b></a> - Logical wait commands can be used to combine multiple queries into a single query. A logical wait block begins with <b>Wait for All</b>, <b>Wait for Any</b> or <b>Wait for Exactly One</b> command, and should end with a <b>Wait for Conditions End</b> command. An <b>All</b> block will wait until all subqueries are true, an <b>Any</b> block will wait until any subquery is true, and an <b>Exactly One</b> block will wait until exactly one subquery is true. Additionally, a logical wait block can be used as a condition for an <a href="#if">if statement</a>. Multiple logical wait blocks can be nested if more complex logic is required. All "Wait for" commands can be used as subqueries, along with the <a href="#role">Set role</a>, <a href="#setplayerweapon">Set player weapon</a>, <a href="#face-towards-like-guard">Face towards like guard</a> and <a href="#set-movement">Set movement type</a> commands, which function as they do when used as an if condition.</li>
```
mrimer, line 364: Changes are looking good!  One more suggested change here to the verbiage, shifting the description toward the new naming.

How about changing the first part to something like:

"Multi-expression wait commands are used to perform an atomic check on multiple conditions, i.e., as a single combined boolean query, which returns a single true or false value. Script execution halts at this point until the multiple conditions are met as specified. The expression block begins..." and so forth.

mrimer (approved):

PR #435, opened 2022-02-02
Make Slayer abandon broken or missing orb or plate.

Hypexion:
When in the door opening state, Slayers will continue to move towards broken orbs, activated one-use plates, and places that used to have something that could open a door, but now do not. This behavior is inconsistent with Halph, who will abandon missing or unusable goals. It's also doesn't make any sense.

Slayers moving to door-openers will now run `CSlayer::ConfirmGoal()`, which checks if there's something at the goal which will be useful for opening a door. If there's nothing there, the current path will be rejected.

This requires one extra function in `CCoordStack` to learn the end of the queue without removing it.

Related thread: http://forum.caravelgames.com/viewtopic.php?TopicID=45542

### DIFF
--- a/DRODLib/Slayer.cpp
+++ b/DRODLib/Slayer.cpp
@@ -651,10 +651,10 @@ bool CSlayer::ConfirmGoal()
 
 	CDbRoom& room = *(this->pCurrentGame->pRoom);
 
-	if (room.GetOSquare(wGoalX, wGoalY) == T_PRESSPLATE && platesToDepress.has(CCoord(wGoalX, wGoalY))) {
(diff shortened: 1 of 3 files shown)


DRODLib/Slayer.cpp
```diff
@@ -941,7 +987,7 @@ void CSlayer::MoveToOpenDoor(CCueEvents &CueEvents)     //(in/out)
 	}
 
 	//Confirm path to goal is still open.
-	if (!ConfirmPath())
+	if (!(ConfirmGoal() &&ConfirmPath()))
```
mrimer, line 990: Add space after &&

PR #438, opened 2022-03-20
Add scripting support to alter room environment

Hypexion:
### Set Lighting commands
Three new script commands that change the light tiles of a room. Supports darkness, ceiling lights and wall lights. In order to make this work with undo and time travel, there's some infrastructure in `RoomWidget` to track when changes to ceiling lights happen, so that they can also unhappen when appropriate.

Names of the light colours are my best guess as to what they are, as they doesn't appear to be an official list.

This might be what is meant be "Set darkness" in #71.

### Weather vars
Five new predefined variables to change (and also query) the state of room weather. Names should be self-explanatory. `_RoomWeather` compacts all the binary weather settings into a single var for simplicity. There's a function in `RoomWidget` to keep weather up to date. Doesn't support changing the cloud images, although such a mechanic is possible.

See thread: http://forum.caravelgames.com/viewtopic.php?TopicID=44130


DRODLib/CurrentGame.cpp
```diff
@@ -68,6 +68,12 @@
 
 #define NO_CHECKPOINT (static_cast<UINT>(-1))
 
+//Max weather values. One less than defines in DROD/RoomWidget.h
+#define MAX_ROOM_LIGHT (6)
```
mrimer, line 72: I'll approve and merge this PR, but for ease of future maintainability, I'd suggest placing these defines in CurrentGame.h and having the defines in DROD/RoomWidget.h based off of these values so they only need to be changed in one place, and so future changes to RoomWidget.h don't cause bugs (iirc, those could be changed with impunity up to this point, if we ever wanted to).

mrimer (approved):

PR #439, opened 2022-03-20
Changes to Wait for weapon functionality

Hypexion:
It probably makes more sense for `Wait for weapon` to be able to check more than one square at a time.
It is more useful if the command ignores the character's own weapon. As it is simple to determine the position of a characters weapon from it's position and orientation, it should not cause a loss of functionality.


mrimer:
I had to go back and confirm that Wait For Weapon is a new 5.2 script command. This change makes sense and seems fine.

mrimer (approved):

PR #443, opened 2022-06-16
Make brained constructs path around oremites

Hypexion:
Thread: http://forum.caravelgames.com/viewtopic.php?TopicID=44966

Currently, constructs use the `GROUND_AND_SHALLOW_WATER` movement type, and it's accompaning brain pathmap. This pathmap tries to make them walk through oremites, which they can't do. This leads to weird situations where a construct gets caught in a dead end due to oremites, even if another path is available. People don't like this.

With a new movement type, `GROUND_AND_SHALLOW_WATER_NO_OREMITES`, constructs can have their own special pathmap that avoids oremites.

People also thought constructs should at least be able to get near the player, even if oremites are in the way. By adding a new `DMASK_SEMI`, tiles in a room can be marked as "semi-obstacles". A tweak to the pathmap construction makes it so moving into a "semi-obstacle" is allowed, but very expensive (1000 addition tiles). This is similar to how monsters using the `WALL` and `WATER` movement types act.

### DIFF
--- a/DRODLib/Construct.h
+++ b/DRODLib/Construct.h
@@ -35,7 +35,7 @@ class CConstruct : public CRockGolem
 {
 public:
 	CConstruct(CCurrentGame *pSetCurrentGame = NULL)
-		: CRockGolem(pSetCurrentGame, M_CONSTRUCT)
+		: CRockGolem(pSetCurrentGame, M_CONSTRUCT, GROUND_AND_SHALLOW_WATER_NO_OREMITES)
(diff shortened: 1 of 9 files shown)


mrimer:
Thanks for the fix! Would you please rebase to resolve the conflict, and then I'll merge?

Hypexion:
> Thanks for the fix! Would you please rebase to resolve the conflict, and then I'll merge?

Conflict is resolved.

mrimer (approved):

PR #447, opened 2022-08-31
Track all image overlay turn counts with player turns

Hypexion:
Image overlays using `DisplayTurns` use the player turn to decide when to end. However, the end turn is set based on the total turn number. Game actions that cause the turn numbers to diverge will extend the duration of image overlays incorrectly. This is a regression introduced by #228.

To fix this issue, image overlays should do all turn tracking based on the player turn.

Related thread: https://forum.caravelgames.com/viewtopic.php?TopicID=45702


Hypexion:
On further consideration, as this issue is a non-gameplay regression, it should probably be in the 5.1.1 branch. Closing in favour of #449.

PR #455, opened 2022-11-07
Additional Scripting Features

Hypexion:
## Introduction
This is a collection of many script features, collated into a single PR to make adding the require strings to mid files easier to manage. There are new script primitives, pre-defined variables, behaviours and commands.

### Moods
The `Talking` and `Dying` moods have been added as options in the speech command.

### Pre-defined variables
`_SpawnCycle`: Gets the current spawn cycle turn, from between 0 and 29.
`_SpawnCycleFast`: For working with speed potions, this gets a spawn cycle value between 0 and 59. Even values are normal turns, odd values are half turns.

### Script primitives
`_EntityWeapon`: Gets the weapon of the armed entity at the given room position. -1 is returned if there isn't someone holding a weapon at that position.
`_BrainScore`: Uses a brain pathmap to return how far a tile is from the player for a given movement type. -1 is returned for tiles that can't reach the player.

### Get natural target modes
`Brained Monster`: Returns the tile that a brained monster would want to move to. This mode takes in to account all the things that would affect a regular brained monster, such as the player hiding or not being a target, and the presence of decoys. In the case where a monster would not do a brained move, it will fall back to finding the tile prefered by the character's movement IQ.
`Best Brain Tile`: Returns the adjacent tile with the best brain score for the character's movement type. If there's no path to the player, it will return the monster's tile.
`Best Brain Direction`: Returns the direction of the tile with the best brain score. Since `Get natural target` returns two values, the `_ReturnX` is the direction accounting for brain-invisible obstacles, and the `_ReturnY` is the best disregarding those obstacles.

[Related thread](https://forum.caravelgames.com/viewtopic.php?TopicID=45737) for brain moves.

### Behaviours
`Use Tunnels`: Allows the character to move through tunnels if stepping in the appropriate direction.
`Instant Movement`: The character won't stop executing commands after a single-step movement. Multiple-step movements still end the turn for both technical and avoiding confusion reasons. [Thread link](https://forum.caravelgames.com/viewtopic.php?TopicID=45467).

### Commands
`Set entity weapon`: Sets the weapon of an armed monster at a specified tile to the specificed weapon. Includes removing it. Doesn't work on the player. [Thread link](https://forum.caravelgames.com/viewtopic.php?TopicID=45412)
`Wait for expression`: Waits until a given expression is equal to a specified value, which isn't simple if you're only using `Wait until var`.
`Link Orb`: Connects an orb or pressure plate to a door, arrow or fire trap, or edits an existing connection. This required some new UI in the editor to allow selection of two target tiles. To make things easy for users, the order of target tiles doesn't matter, as long as one is an orb/plate and the other is an orb target.

There are also three new commands for looking at build markers. One to wait for build markers, and two to check for specific build market types. [Thread link ](https://forum.caravelgames.com/viewtopic.php?TopicID=45589)


DROD/EditRoomScreen.cpp
```diff
@@ -7170,6 +7192,13 @@ void CEditRoomScreen::SetSignTextToCurrentRoom(
 	case ES_GETSQUARE:
 		wstrSignText += g_pTheDB->GetMessageText(MID_GetRoomSquare);
 		break;
+	case ES_GETSQUARES1:
+	case ES_GETSQUARES2:
+		wstrSignText += g_pTheDB->GetMessageText(MID_GetRoomSquare);
+		wstrSignText += L" (";
```
mrimer, line 7198: I think we need to use vars/chars defined in wchar.* for these unicode symbols so this works on other architectures/OSes.

DRODLib/Character.cpp
```diff
@@ -3559,6 +3760,86 @@ void CCharacter::BuildTiles(const CCharacterCommand& command, CCueEvents& CueEve
...
+	COrbData* orb;
+
+	if (room.GetTSquare(px, py) == T_ORB) {
+		orb = room.GetOrbAtCoords(px, py);
+		linkX = pw;
+		linkY = ph;
+		if (!orb) {
+			orb = room.AddOrbToSquare(px, py);
+		}
+	}	else if (room.GetOSquare(px, py) == T_PRESSPLATE)	{
```
mrimer, line 3784: Fix inconsistent whitespacing here and below.

mrimer (commented):
Beautiful work!

Ready to merge after minor changes in a couple places.

mrimer (approved):

PR #458, opened 2022-11-17
Rework of Brained Construct Pathing

Hypexion:
There was [some discussion](https://forum.caravelgames.com/viewtopic.php?TopicID=44966&page=0#444706) on how Constructs that are brained could act. A proposal is that if no non-oremite paths are available, a Construct will move towards oremites closest to the player. This makes them seem smarter.

Implementing this was a little tricker, as it requires a secondary pathmap which can path over oremites normally to provide scores. When a semi-obstacle tile is encountered, its score is made equal to its score on the sub-pathmap multiplied by 1000.
Due to memory issues, the only way I was able to have the subpath system work was to add an option for a `CPathMap` to own another `CPathMap`, which it can pass tile mask data to, with alterations if required. Since this is only required for one movement type at the moment, the memory and computation overhead is not too much.

There is also a fix, as adding the new movement type in #443 ended up removing Constructs' ability to move over shallow water tiles. By default, wading is now possible for all `GROUND_AND_SHALLOW_WATER` movement variants.

### DIFF
--- a/DRODLib/DbRooms.cpp
+++ b/DRODLib/DbRooms.cpp
@@ -2337,7 +2337,8 @@ void CDbRoom::CreatePathMap(
 		if (eMovement == WALL || eMovement == WATER) //allow for non-wall/water
 			dwPathThroughObstacleCost = 1000;         //squares to be in the path
 		this->pPathMap[eMovement] = new CPathMap(this->wRoomCols, this->wRoomRows,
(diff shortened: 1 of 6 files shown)


mrimer:
I've merged PR 457.

mrimer (approved):
I'm pleasantly surprised how clean this change turned out to be in your implementation. Thanks!

PR #463, opened 2022-12-11
Customise scoring

kieranmillar:


kieranmillar:
Whoops

kieranmillar:
There I fixed it. Accidentally pointed it at master, not rpg-1.3-dev.

This PR is best explained by this forum thread. Includes a screenshot of how it looks:
https://forum.caravelgames.com/viewtopic.php?TopicID=44777

mrimer (approved):
Thank you for the PR! Very careful and thorough changes.
I like this addition.

mrimer:
In addition to patching the new MID texts in DbBase.cpp, we also need to add them to a Texts .uni file (probably the one where other script vars are) so they will be added to the .dat file on next .dat file rebuild.

PR #474, opened 2022-12-27
Upgrade SDL2 and SDL2 TTF dependencies and accept alternative arrow keybinds

kieranmillar:
Issue was reported that DROD does not recognise numpad keys 8, 6, 4 and 2 when numlock is off, but the others are fine.
https://forum.caravelgames.com/viewtopic.php?TopicID=45829
At first I believed this to be an SDL2 issue, and so upgraded SDL2 and SDL2_ttf dependendencies to their latest stable versions. However, this did not seem to fix the issue. I left in the upgrade anyway, because I put in the work.

Instead, I have added functionality to recognise the arrow keys as alternative key binds for the 4 numpad keys with an issue. If you press a key and it is not initially mapped to a game command, the game will then check an alternative keybind mapping with the arrow keys and num keys flipped to see if there is then a match.

P.S. SDL2_ttf no longer uses libfreetype-6.dll so this no longer needs to be included.

(I do not know why this commit list has 2 years of merging with the latest version of master listed. Oh well.)

### DIFF
--- a/CompilingDrod_Win.md
+++ b/CompilingDrod_Win.md
@@ -39,8 +39,8 @@ The game requires the following libraries. Different versions may potentially be
  -  `libvorbis-1.3.3` => http://downloads.xiph.org/releases/vorbis/libvorbis-1.3.5.tar.gz
  -  `lpng-1512` => https://downloads.sourceforge.net/project/libpng/libpng15/older-releases/1.5.12/lpng1512.zip
  -  `metakit-2.4.9.7` => https://github.com/jnorthrup/metakit/archive/master.zip
- -  `sdl2-2.0.12` => https://www.libsdl.org/release/SDL2-2.0.12.zip
- -  `sdl-ttf-2.0.14` => https://www.libsdl.org/projects/SDL_ttf/release/SDL2_ttf-devel-2.0.14-VC.zip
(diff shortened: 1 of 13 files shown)


Hypexion:
You can probably do a "squash" rebase to combine all your commits into one. Since they're just merges it should be fine.

We should check that this doesn't cause any issues with the new control mapping system in 5.2

Tahnan:
. o O ( "Issue was reported"?  I have a *name*, you know. )

Tahnan:
(no, but seriously, thanks for being on top of this! I'd help out, but C++ is an alien landscape for me.)

kieranmillar:
> We should check that this doesn't cause any issues with the new control mapping system in 5.2

I took a look at the 5.2 branch. The key to text string mappings are already done in GameConstants, except even more comprehensively. The keys are now passed around RoomScreen not as direct keycodes but in an InputKey wrapper to allow modifier keys like Shift and Ctrl to work too. So the changes there will need some tweaks but ultimately the main feature of letting the numpad numbers and arrows be used interchangeably can still be carried over.

mrimer:
Thank you for the upgrades and fixes, @kieranmillar !

These changes look good to me.  I agree it would make sense to rebase these commits, squashing all the commits into one, prior to merging. Please ask for help if needed.

kieranmillar:
I tried to rebase the commits but instead made it worse, so instead I have just done the PR all over again, see https://github.com/CaravelGames/drod/pull/483

PR #475, opened 2022-12-27
Solution and project files for VS 2022

kieranmillar:
Solution and project files for VS 2022, that use the 2022 platform toolset.

I'll do DROD PRG in a separate PR so it can be applied to the rpg branch.


kieranmillar:
Going to remove this, I don't think the code base is stable enough yet to warrant more solution files at this time.

PR #478, opened 2023-02-14
Fix cast int->unsigned int

yyweiss:
g++ 12 creates a compile error on narrowing the int.


mrimer:
Thank you, @yyweiss!

I'd like to request someone test this change on a non-g++ compiler to confirm it's non-impactful.

PR #480, opened 2023-04-02
Add two more special commands

Hypexion:
It's two more special commands. Along with a few changes to manage there now being more than one.

Thread: http://forum.caravelgames.com/viewtopic.php?TopicID=45873

### DIFF
--- a/DRODLib/Character.cpp
+++ b/DRODLib/Character.cpp
@@ -4573,13 +4573,16 @@ bool CCharacter::DidPlayerInput(
 	getCommandX(command, px);
 
 	//CMD_EXEC_COMMAND is translated to CMD_WAIT, so we need to kludge handling here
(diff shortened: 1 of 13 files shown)


DRODLib/Character.cpp
```diff
@@ -4577,6 +4577,10 @@ bool CCharacter::DidPlayerInput(
 		return false;  //nLastCommand is a false "wait"
 	if (px == CMD_EXEC_COMMAND) {
 		return CueEvents.HasOccurred(CID_CommandKeyPressed); //only way to detect
+	} else if (px == CMD_EXEC_COMMAND_TWO) {
```
mrimer, line 4580: Nit: don't need the "else", since there's a return in the then-block. I think it reads cleaner this way.

mrimer (approved):

mrimer (approved):

PR #482, opened 2023-05-15
RPG: Additional Scripting Features

Hypexion:
## Introduction
This is a group of new scripting features for RPG.

### Pre-defined vars
New vars to access various game state information

`_RoomX`, `_RoomY`: The position of the current room relative to the level entrance (read-only)
`_LevelMultiplier`: Read-only access to the current level's base multiplier value
`_MyDescription`: A NPC variable to allow additional tooltip content ([Thread](https://forum.caravelgames.com/viewtopic.php?TopicID=45811))

### Behaviour
A new NPC behaviour

`Removes Sword` - Equipment behaviour that causes the player not to have a sword. ([Thread](https://forum.caravelgames.com/viewtopic.php?TopicID=26959))

### Custom processing sequence
All the stuff has been plumbed in to allow NPCs to have a custom processing sequence. Doesn't effect the order equipment scripts get run because those are run in a special way.


mrimer (commented):
One syntactic question; otherwise, approved!

mrimer (approved):

PR #486, opened 2023-05-18
Divide script help into two files

Hypexion:
In the 5.2 branch, I've occationally had DROD crash when attempting to view the `script.html` file via the in-game browser. It seems to be caused by SDL not being able to make a surface once the length of the file goes beyond a certain point.

For lack of a better solution, I've separated out the information about _MyScript variable injection into a new help file, called `myscript.html`.

Some issues with the html in the help file have also been corrected.


mrimer (approved):
Ah, yes, it's true that surfaces can only be so large. I've run into that issue as well for the sprite sheets.

PR #487, opened 2023-05-18
RPG: Add prompt when starting new game if a game is active

Hypexion:
Starting a new game in DROD RPG overwrites the existing game, in a non-recoverable manner. This can lead to a loss of progress if a manual save has not been created recently.

To reduce the chance of this, we can confirm if the player really wants to start a new game with a yes/no dialog. This prompt can be disabled in the settings, if the player wants.

Thread: https://forum.caravelgames.com/viewtopic.php?TopicID=45874


Hypexion:
Converting to draft since this needs strings set up.

mrimer (approved):

PR #491, opened 2023-05-22
Enlarge Add Command dialog

Hypexion:
Compared to the available screen space, the dialog for adding script commands is quite small. This can be fixed by making that dialog larger:
image

Since the dialog is being enlarged, the UI for variable-related commands has been rearranged, to move the value input box and increase its size. This is helpful when writing assignments using script primitives:
image

Behind the scenes, the third on/off listbox has been upgraded to be dynamically positioned, as its fixed position no longer worked for all commands that used it. This allowed the fourth on/off listbox to be removed, as it only existed to have a control at its specific location. (Possibly more widgets could become dynamically placed but that is outside of the scope for this PR.)

### DIFF
--- a/DROD/CharacterDialogWidget.cpp
+++ b/DROD/CharacterDialogWidget.cpp
@@ -922,8 +922,8 @@ void CCharacterDialogWidget::AddCommandDialog()
 //Create another dialog for entering a script command.
 //Separated from constructor for readability.
 {
-	static const UINT CY_COMMAND_DIALOG = 412;
-	static const UINT CY_ACTIONLISTBOX = 13*LIST_LINE_HEIGHT+4;
+	static const UINT CY_COMMAND_DIALOG = 650;
+	static const UINT CY_ACTIONLISTBOX = 23*LIST_LINE_HEIGHT+4;
 
 #ifdef RUSSIAN_BUILD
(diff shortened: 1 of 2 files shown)


DROD/CharacterDialogWidget.cpp
```diff
@@ -1326,7 +1328,13 @@ void CCharacterDialogWidget::AddCommandDialog()
 	static const int X_IMAGEOVERLAYTEXT = X_IMAGELABEL;
 	static const int Y_IMAGEOVERLAYTEXT = Y_IMAGELABEL + CY_IMAGELABEL;
 	static const UINT CX_IMAGEOVERLAYTEXT = CX_TEXT;
-	static const UINT CY_IMAGEOVERLAYTEXT = 284;
+	static const UINT CY_IMAGEOVERLAYTEXT = CY_ACTIONLISTBOX;
+
+	//Positions for on off box 3
+	onOffListBox3Positions.insert({ CCharacterCommand::CC_SetNPCAppearance, {KEEPBEHAVIORLABEL_X, Y_ONOFFLISTBOX} });
```
mrimer, line 1334: I think this is c++11 syntax.
Suggest using make_pair.

mrimer (commented):
One syntactic thing to change. Approved after this change.

mrimer (approved):

PR #493, opened 2023-05-26
Fix Ctrl-Wait sometimes waiting for 60 turns

Hypexion:
Keyboard inputs are weird, and this is somehow causing some players to wait 60 turns when they do a Ctrl-Wait instead of 30 turns. This is because two input events are registered by SDL, which causes a macro'd wait command to be sent to the engine twice.

Luckily `FrontEndLib` can tell when a key event is repeated. When a repeated keypress is seen, we can increment its `repeat` value by one. Then in `CGameScreen::OnKeyDown` if we see a macro'd wait command with exactly one repeat, we do not pass it to the engine. This solves the problem.

### DIFF
--- a/DROD/GameScreen.cpp
+++ b/DROD/GameScreen.cpp
@@ -1553,6 +1553,10 @@ void CGameScreen::OnKeyDown(
 	}
 	if (nCommand != CMD_UNSPECIFIED)
 	{
+		//Sometimes, SDL will send the same input twice. If this happens for long waits, we don't
+		//want two of them to happen. Repeated inputs are counted, so for the first repeat, a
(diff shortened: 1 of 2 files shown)


DROD/GameScreen.cpp
```diff
@@ -1553,6 +1553,10 @@ void CGameScreen::OnKeyDown(
 	}
 	if (nCommand != CMD_UNSPECIFIED)
 	{
+		if (nCommand == CMD_WAIT && bMacro && Key.repeat == 1) {
```
mrimer, line 1556: Suggest adding a comment to explain how this logic helps, otherwise this looks unintuitive.

mrimer (approved):

PR #499, opened 2023-06-03
RPG: Add item bonus preview effect.

mrimer:
Like the enemy damage preview effect, this change adds a persisting numeric display for pending item bonuses (i.e., health, att/def gems).

### DIFF
--- a/drodrpg/DROD/DamagePreviewEffect.cpp
+++ b/drodrpg/DROD/DamagePreviewEffect.cpp
@@ -36,19 +36,171 @@
 const int YOFFSET = 3;   //this font draws a bit too low
 
 //********************************************************************************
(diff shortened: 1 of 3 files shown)


Hypexion:
This looks like it's replacing the existing damage preview effect. Did you mean to push some different files?

mrimer:
> This looks like it's replacing the existing damage preview effect. Did you mean to push some different files? EDIT: No it's just Github showing things strangely.

I added a new base class to the damage preview effect.

Hypexion:
> > This looks like it's replacing the existing damage preview effect. Did you mean to push some different files? EDIT: No it's just Github showing things strangely.
> 
> I added a new base class to the damage preview effect.

Ah, it was a case of Github displaying things strangely. This all looks good!

PR #513, opened 2023-06-10
Remove grid style option for puzzle mode

Hypexion:
The grid style option doesn't work because it needs graphics, but none were ever supplied. It is also unclear what "grid style" is meant to be, as the comment for in it `GridEffect` says "size of the grid". Due to this is it is probably best to remove it. (The strings have to remain to preserve the MID ordering)

Thread: https://forum.caravelgames.com/viewtopic.php?TopicID=44932

### DIFF
--- a/DROD/GridEffect.cpp
+++ b/DROD/GridEffect.cpp
@@ -33,13 +33,10 @@
 
 #include <BackEndLib/Assert.h>
 
-const UINT CGridEffect::wGridStylesCount = 4;
-
 //********************************************************************************
 CGridEffect::CGridEffect(
 //Params:
 	CWidget *pSetWidget,    //(in)   Should be a room widget.
-	const UINT wGridStyle,  //(in)   Size of the grid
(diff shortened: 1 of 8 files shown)


mrimer:
@Hypexion , I think skell actually prepared new tile sprites. I didn't include them in the patch I posted on the forum because I had forgotten that new tiles were added for 5.2. I don't want to yank the feature simply because I didn't include the tiles. If I'm missing something in regards to this change, however, please let me know.

Hypexion:
> @Hypexion , I think skell actually prepared new tile sprites. I didn't include them in the patch I posted on the forum because I had forgotten that new tiles were added for 5.2. I don't want to yank the feature simply because I didn't include the tiles. If I'm missing something in regards to this change, however, please let me know.

If they are actually new graphics that can be included then this PR shouldn't be required.

Hypexion:
Since there are graphics, closing.

PR #527, opened 2023-07-01
Additional bug fixes

Hypexion:
There were some problem with recent bug fixes, which require some more fixing:

- Subtitle effect starting asserting whenever `CSubtitleEffect::Update` was called. This is due to an issue with changing opacity. The opacity changes should be skipped if the text duration is zero. (Thread: http://forum.caravelgames.com/viewtopic.php?TopicID=45938)
- Slayer starting rejecting all orbs. This is due to the goal tile being checked instead of the orb tile, leading to incorrect assessment. If the orb tile is checked, things work correctly. (Thread: http://forum.caravelgames.com/viewtopic.php?TopicID=45542)
- The new key binding system forbids any command from sharing a key code. However, some default key configurations conflict. New defaults have been set to prevent conflicts. (No thread)

### DIFF
--- a/DRODLib/GameConstants.cpp
+++ b/DRODLib/GameConstants.cpp
@@ -93,7 +93,7 @@ namespace InputCommands
 		keyDefinitions[DCMD_PuzzleModeOptions] = new KeyDefinition(CMD_EXTRA_PUZZLE_MODE_OPTIONS, "Key_PuzzleModeOptions", MID_Command_PuzzleModeOptions, BuildInputKey(SDLK_F3, false, true, false));
 		keyDefinitions[DCMD_ToggleTurnCount] = new KeyDefinition(CMD_EXTRA_TOGGLE_TURN_COUNT, "Key_ToggleTurnCount", MID_Command_ToggleTurnCount, SDLK_F7);
 		keyDefinitions[DCMD_QuickDemoRecord] = new KeyDefinition(CMD_EXTRA_QUICK_DEMO_RECORD, "Key_QuickDemoRecord", MID_Command_QuickDemoRecord, SDLK_F4);
-		keyDefinitions[DCMD_ToggleDemoRecord] = new KeyDefinition(CMD_EXTRA_TOGGLE_DEMO_RECORD, "Key_ToggleDemoRecord", MID_Command_ToggleDemoRecord, SDLK_F5);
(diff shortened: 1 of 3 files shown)


DRODLib/GameConstants.cpp
```diff
@@ -93,7 +93,7 @@ namespace InputCommands
 		keyDefinitions[DCMD_PuzzleModeOptions] = new KeyDefinition(CMD_EXTRA_PUZZLE_MODE_OPTIONS, "Key_PuzzleModeOptions", MID_Command_PuzzleModeOptions, BuildInputKey(SDLK_F3, false, true, false));
 		keyDefinitions[DCMD_ToggleTurnCount] = new KeyDefinition(CMD_EXTRA_TOGGLE_TURN_COUNT, "Key_ToggleTurnCount", MID_Command_ToggleTurnCount, SDLK_F7);
 		keyDefinitions[DCMD_QuickDemoRecord] = new KeyDefinition(CMD_EXTRA_QUICK_DEMO_RECORD, "Key_QuickDemoRecord", MID_Command_QuickDemoRecord, SDLK_F4);
-		keyDefinitions[DCMD_ToggleDemoRecord] = new KeyDefinition(CMD_EXTRA_TOGGLE_DEMO_RECORD, "Key_ToggleDemoRecord", MID_Command_ToggleDemoRecord, SDLK_F5);
+		keyDefinitions[DCMD_ToggleDemoRecord] = new KeyDefinition(CMD_EXTRA_TOGGLE_DEMO_RECORD, "Key_ToggleDemoRecord", MID_Command_ToggleDemoRecord, BuildInputKey(SDLK_F6, false, false, true));
```
mrimer, line 96: Why are we shifting from F5 to F6?

Hypexion: It's a side effect of allowing these keys to be rebinded. Due to how input -> command lookup works, if multiple commands share a key, only one will be recognised. The player can't put themselves in that situation, but some of the default keys collide.

There is probably a clever way to allow non-clashing commands to share keys, but this is the simple fix.

mrimer: Thank you. This makes sense.

I'd expect we need to update the help doc that mentions this key/case. Would you please check and make that change?

mrimer (approved):

PR #541, opened 2023-09-08
Add script command to reset _MyScript variables

Hypexion:
Simple command that sets the value of all _Myscript variables to `NO_OVERRIDE`. Makes things easier for scripters and is probably faster than using `Set Var`.

Thread: https://forum.caravelgames.com/viewtopic.php?TopicID=45934

### DIFF
--- a/DRODLib/Character.cpp
+++ b/DRODLib/Character.cpp
@@ -3571,6 +3571,12 @@ void CCharacter::Process(
 			}
 			break;
 
+			case CCharacterCommand::CC_ResetOverrides: {
+				bProcessNextCommand = true;
+				paramX = paramY = paramW = paramH = paramF = NO_OVERRIDE;
+			}
+			break;
+
 			//Deprecated commands
 			case CCharacterCommand::CC_GotoIf:
 			case CCharacterCommand::CC_WaitForHalph:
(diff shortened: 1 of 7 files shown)


mrimer (approved):

DRODLib/Character.cpp
```diff
@@ -3571,6 +3571,11 @@ void CCharacter::Process(
 			}
 			break;
 
+			case CCharacterCommand::CC_ResetOverrides: {
+				paramX = paramY = paramW = paramH = paramF = NO_OVERRIDE;
+			}
```
mrimer, line 3576: I think we need
bProcessNextCommand = true;

Hypexion: So we do. I've pushed a new commit to fix

Texts/Speech.uni
```diff
@@ -3285,3 +3285,7 @@ Dying
 [MID_Talking]
 [eng]
 Talking
+
+[MID_ResetOverrides]
+[eng]
+Reset _Myscript variables
```
mrimer, line 3291: _Myscript -> _MyScript

mrimer (approved):

PR #557, opened 2023-10-21
RPG: Make shields provide doubled DEF against weak monsters

Hypexion:
Makes weakness behaviours and `_MyWeakness` functional on shield items. The player's effective defense is doubled against monsters their shield is strong against (unless that monster has `No enemy defense`). This applies during combat, and when taking damage from monsters that attack outside of combat.

Thread: https://forum.caravelgames.com/viewtopic.php?TopicID=46014


drodrpg/Data/Help/1/script.html
```diff
@@ -549,10 +549,10 @@ <h3><u>Character Script Commands</u></h3>
       NPC will generate roach eggs or another monster type when the player engages another monster in combat. The type of monster generated can be controlled by the <b>_MySpawn</b> and <b>_QueenSpawn variables</b>.</li>  
   <li><a name="be_goblin"><u>Goblin Weakness</u></a>:
 	  NPC is weak to goblin blades (i.e. player gets x2 ATK).
-	  Embues custom weapons with goblin blade attribute.</li>
+	  Embues custom equipment with goblin blade attribute. Shields with goblin blade attribue instead provide x2 DEF against attacks from goblin enemies.</li>
```
mrimer, line 552: attribue -> attribute
(also below)

mrimer (approved):

PR #558, opened 2023-10-24
Update SDL_ttf to version 2.20.1

Hypexion:
Recent discoveries make it possible to update the development branches to version `2.20.1` of SDL_ttf. A code fix is required, along with a change to one font file. This solves issues that occured during a previous upgrade attempt.

The first issue was that text outlines became displaced. This is probably due to a fix in version of `2.0.15` SDL_ttf related to outline positions. `CFontManager::RenderWord` appears to contain a work around for this problem, which has been removed as it is no longer needed.

The second issue is that some text was displaced, and text in the title menu was cut-off. This is due to a fix in the freetype library, where it began respecting the `useTypoMetrics` flag present in TrueType fonts. The version of the Tom's New Roman font included with DROD had this flag active, leading to a number of font values changing, which ultimately broke all the font drawing. A later version of this font, available from Tom [here](http://fonts.tom7.com/fonts93.html) does not have this flag set, and therefore does not make everything ugly or unreadable.

Since fonts can be tricky, a fonts section has been added to the readme.


mrimer (commented):
Suggest we also need to include the new version of the font file itself in place of the old one (in Data/Fonts) with this PR?

mrimer:
I've updated all my copies of the font file to the newer version for future builds.

mrimer (approved):

PR #560, opened 2023-10-27
Count entity type and Count item script commands

Hypexion:
Script commands for counting the amount of things in a room region.

This required a bit of refactoring for the tile counting command, shifting the actual check for a square into the new `CDbRoom::DoesSquareContainTile`. As a side effect, this fixes a bug where variant orbs and plates would not be seen by `Wait for Item` in certain situations.

Thread: https://forum.caravelgames.com/viewtopic.php?TopicID=46007

### DIFF
--- a/DRODLib/Character.cpp
+++ b/DRODLib/Character.cpp
@@ -4360,7 +4360,6 @@ bool CCharacter::IsTileAt(const CCharacterCommand& command) const
 		return false;
 
 	const bool bFakeElement = bIsFakeElementType(pflags);
-	const bool bLightingElement = bIsLighting(pflags);
 	const UINT tile = bFakeElement ? bConvertFakeElement(pflags) : pflags;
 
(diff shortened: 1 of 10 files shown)


DRODLib/Character.cpp
```diff
@@ -4289,6 +4309,61 @@ bool CCharacter::IsEntityTypeAt(
...
+			if (!pMonster) {
+				continue;
+			}
+
+			if (pMonster->wType == pflags) {
+				++count;
+				continue;
+			}
+
+			switch (wType) {
```
mrimer, line 4342: I'm unclear of the origin of this 'wType' var. Same for the instance below.

Hypexion: Thanks for catching. Accidentally used the local version of `wType` instead of `pMonster->wType` the first time and `pflags` the second time.

mrimer (approved):

PR #563, opened 2023-12-08
Fix interaction when pushing character off of falling tile

Hypexion:
If a falling tile such as a trapdoor drops by having an monster pushed off of it, it is possible for the pushing entity to end up on an otherwise impassable pit or water tile. This is inconsistent with other situations where a tile is altered to pit or water beneath an entity, which can lead to them dropping or drowning.

To fix, monsters now set a flag `bPushedOtherMonster` when they body push a monster. If this flag is set after their movement is processed, monsters that can push other monsters will call `CDbRoom::CheckForFallingAt` at their new position. This will cause them to drop if their destination tile is unsuitable. `CCurrentGame::ProcessPlayer` has had a similar change made to allow the same for player movement.

### DIFF
--- a/DRODLib/Monster.cpp
+++ b/DRODLib/Monster.cpp
@@ -134,6 +134,7 @@ void CMonster::Clear()
 	this->stunned = 0;
 	this->bNewStun = false;
 	this->bPushedThisTurn = false;
+	this->bPushedOtherMonster = false;
 	this->ExtraVars.Clear();
 	while (this->Pieces.size())
 	{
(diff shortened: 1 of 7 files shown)


DRODLib/Monster.cpp
```diff
@@ -105,7 +105,7 @@ CMonster::CMonster(
 	, bAlive(true)
 	, bForceWeaponAttack(false)
 	, stunned(0), bNewStun(false)
-	, bPushedThisTurn(false)
+	, bPushedThisTurn(false), bPushedOtherMonster(false)
```
mrimer, line 108: Suggest we reset this var in CMonster::Clear().

mrimer (approved):

PR #564, opened 2023-12-22
Adding build support and instructions for Raspberry Pi 4

mpettitt:
Adding support for Arm64 (aarch64) architecture, as used by Raspberry Pi 4 (and possibly 5 - don't have one to check), based on comments from https://forum.caravelgames.com/viewtopic.php?TopicID=45518 and https://forum.caravelgames.com/viewtopic.php?TopicID=45992


mrimer (approved):
Thank you for adding ARM/Pi support!

PR #565, opened 2024-01-01
RPG: Add Thin ice and Disabled arrows elements

Hypexion:


mrimer (approved):
Thank you so much for migrating these features!

PR #570, opened 2024-02-21
RPG: Add DEF nullifying tile (mist)

Hypexion:
Adds a new transparent layer tile (mist) that nullifies the DEF stat of a player or monster standing on it. The tile takes some mechanic cues from Fluff, being destroyed when doors are closed under it, and preventing briar growth.

Crates and mirrors can be pushed onto mist, and mist does not affect a player standing on a crate. To support this graphically, covered T-layer objects are now drawn into a room (as in DROD TSS), and are included in right-click tooltips.

As mist is transparent and uses corner filler when drawn, it requires its own function `CRoomWidget::DrawMistTile` for drawing. Its tile image calculations are also placed in a new function due to the requirement to check both the T-layer and covered T-layer object during calculation.


mrimer (approved):
Thank you for adding these new features!

How are serpents handled?

Hypexion:
> Thank you for adding these new features!
> 
> How are serpents handled?

For the moment, serpents are debuffed if their head is on a mist tile, and their tails aren't independently handled. Having tails be affected by mist on their tail probably requires some work in the combat system to better handle multi-part monsters.

PR #571, opened 2024-02-22
RPG: More robust prevention of infinite egg laying

Hypexion:
If multiple egg-laying monsters generate different monsters, the checks to stop spawned monsters triggering additional monster may not work, allow for the generation of an arbitary number of combatable monsters.

To prevent this, an additional persisted flag, `bEggSpawn` can be added to `CMonster`, which indicates the monster was generated by an egg-laying monster. This prevents infinite spawn loops.

The older behavior is still present to preserve backwards-compatibility.


mrimer (commented):
Thank you. When does this flag get reset on a monster? (Or is the variable to always remain 'true' once set?)

Hypexion:
> Thank you. When does this flag get reset on a monster? (Or is the variable to always remain 'true' once set?)

Currently it remains true once set indefinitely, as a monster can't become not spawned by an egg-layer. However that should only apply in terms of foward game state. I've pushed an updated commit to reset it in `CMonster::Clear` to be safe.

mrimer (approved):

PR #573, opened 2024-03-01
Add commands to cap image overlay duration

Hypexion:
Adds two new image overlay commands, `TimeLimit` and `TurnLimit`, that place a restriction on how long an image overlay can live for.

`TimeLimit` causes the overlay to end after the specified amount of time. Primarily useful for capping the length of dynamic effects where it may be difficult to otherwise set the duration.
`TurnLimit` causes the overlay to end after the specified number of turns pass. Useful for placing a turn limit on an effect that loops forever, in a way that is self-contained.

### DIFF
--- a/DROD/ImageOverlayEffect.cpp
+++ b/DROD/ImageOverlayEffect.cpp
@@ -97,9 +97,9 @@ CImageOverlayEffect::CImageOverlayEffect(
 	, repetitions(0), xRepeatOffset(0), yRepeatOffset(0)
 	, index(UINT(-1))
 	, loopIteration(0), maxLoops(0)
-	, startOfNextEffect(dwStartTime)
+	, startOfNextEffect(dwStartTime), endTime(0)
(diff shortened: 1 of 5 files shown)


DROD/ImageOverlayEffect.cpp
```diff
@@ -399,15 +409,19 @@ bool CImageOverlayEffect::IsCurrentCommandFinished() const
...
 		return this->executionState.endTurn == gameTurn;
 	}
 	else
 		return true;
 }
 
+UINT CImageOverlayEffect::GetGameTurn() const
+{
+	const CDbRoom* pRoom = this->pRoomWidget->GetRoom();
+	return pRoom ? pRoom->GetCurrentGame()->wPlayerTurn : 0;
```
mrimer, line 422: We might also need to check whether GetCurrentGame() returns null.

mrimer (approved):

PR #577, opened 2024-03-10
RPG: Add mist vent

Hypexion:
Adds a new tile that generates and expands mist after a monster is battled. A vent with no T-layer item on its square will create a single tile of mist at its location. A vent with mist already on it will expand all connect mist. If multiple vents are attached to a group of connect mist tiles, only one expansion occurs.

There are a few mist-specific methods added, to support mist being able to connect via covered tiles.

This PR also includes a fix for mist drawing when new mist tiles are plotted diagonally adjacent to existing mist.


mrimer (approved):

mrimer:
Thank you, @Hypexion !

PR #578, opened 2024-03-14
Add Set player behavior script command

Hypexion:
Adds a script command that allows behaviours of the player to be overridden, similarly to character behaviours.

## Basic details
There are five states a player behaviour can be set to: `Default`, `On`, `Off`, `Powered` and `Unpowered`. When set to `Default`, the player character's appearance is used to determine the interactions for that behaviour. `On` and `Off` cause a behaviour to always be active or inactive, while `Powered` and `Unpowered` cause the behaviour to be controlled by the power token state.

A new member has been added `CSwordsman` to store behaviour overrides. The `Default` state is represented by the behaviour not being present in the map, which helps with backwards compatibility.

The function `CSwordsman::HasBehavior` has been added to check for player behaviour states. It both returns if the queried behaviour is being overridden, and if the behaviour is active based on the behaviour and game state.

Clones and Temporal Projections inherit player behaviours parallel to how they inherit player roles. Clones always mirror the player's behavior states, while Projections gain all behaviours the player had at the time of the Projection's creation.

## Serialisation
To allow player behaviour changes to be retained between rooms and game sessions, they are saved to a game variable `PlayerBehaviorOverrides`. Currently, each override requires two bytes, one to save the behaviour type and one to save its state. If no overrides are active, nothing is saved. This currently limits the number of behaviour types to 255, but using a three-byte system is possible. For convenience, the data is stored as a string, which means no behaviour can have an index of zero. In a future version, it should be possible to move this information to a proper subrecord in the saved game database.

## Refactors
To allow for granular interaction overrides, a lot of checks for player interactions have been moved into their own functions. This also consolidates a lot of these checks into the actual swordsman object, rather than having them in various other places.

(As a side-effect, tunnel refactors fix this bug: http://forum.caravelgames.com/viewtopic.php?TopicID=46097)

Thread: https://forum.caravelgames.com/viewtopic.php?TopicID=46043


mrimer (approved):

mrimer:
Thank you, @Hypexion !

PR #579, opened 2024-03-14
Fix demo playback ending a turn early

Hypexion:
Thread: https://forum.caravelgames.com/viewtopic.php?TopicID=46088


mrimer (commented):
I believe these vars are unsigned. Not sure about subtracting one, as it could roll over when at zero, potentially introducing regressions elsewhere.

How about
`return this->pCurrentGame->wTurnNo > this->pDemo->wEndTurnNo;`
?

mrimer (approved):

PR #587, opened 2024-05-18
Add support for array script variables

Hypexion:
Adds new functionality to scripts by allowing arrays. A script variable starting with a `#` becomes a global array, and a variable starting with a `@` becomes a local variable. To avoid weird issues with serialization, once a variable is created, it can't be renamed in a way that changes its "arrayness".

Internally, arrays are technically maps, as they can be non-continuous. Negative indexes are also supported, although this is not unheard of in some languages.

Global arrays are serialized back to the packed stats object when saving. The number of active items in the array is saved as the first value, followed by a sequence of key/value pairs. The `readBpUINT` and `writeBpUINT` functions have been moved to `CDbSavedGame` to allow them to reused for this. To save space, values of zero are not packed when serializing (as getting an int value from a map will return zero is no value is assigned to that key).

New script commands have been added to write to arrays. Reading from arrays is handled by expression parsing functions.

Thread: http://forum.caravelgames.com/viewtopic.php?TopicID=46113

### DIFF
--- a/DROD/CharacterDialogWidget.cpp
+++ b/DROD/CharacterDialogWidget.cpp
@@ -185,6 +185,11 @@ const UINT TAG_VARCOMPLIST2 = 873;
 const UINT TAG_ORBAGENTLIST = 872;
 const UINT TAG_PLAYERBEHAVE_LIST = 871;
 const UINT TAG_PLAYERBEHAVESTATE_LIST = 870;
+const UINT TAG_ARRAYVARLIST = 869;
+const UINT TAG_ARRAYVAROPLIST = 868;
+const UINT TAG_ARRAYINDEX_LABEL = 867;
+const UINT TAG_ARRAYVAR_REMOVE = 866;
+const UINT TAG_ARRAYVAR_TEXTLABEL = 865;
 
 const UINT MAX_TEXT_LABEL_SIZE = 100;
 
@@ -328,6 +333,14 @@ void CRenameDialogWidget::OnDoubleClick(
 		}
 		break;
 
+		case TAG_ARRAYVARLIST:
+		{
+			CCharacterDialogWidget* pParent = DYN_CAST(CCharacterDialogWidget*,
+				CWidget*, this->pParent);
+			pParent->RenameVar(true);
+		}
+		break;
+
 		case TAG_CHARACTERLISTBOX:
 		{
 			CCharacterDialogWidget *pParent = DYN_CAST(CCharacterDialogWidget*,
@@ -421,6 +434,7 @@ CCharacterDialogWidget::CCharacterDialogWidget(
 	, pOnOffListBox(NULL), pOnOffListBox2(NULL), pOpenCloseListBox(NULL)
 	, pGotoLabelListBox(NULL), pMusicListBox(NULL)
 	, pVarListBox(NULL), pVarOpListBox(NULL), pVarCompListBox(NULL), pVarCompListBox2(NULL)
+	, pArrayVarListBox(NULL), pArrayVarOpListBox(NULL)
 	, pWaitFlagsListBox(NULL), pImperativeListBox(NULL), pBuildItemsListBox(NULL)
 	, pBuildMarkerListBox(NULL), pWaitForItemsListBox(NULL)
 	, pCharNameText(NULL), pCharListBox(NULL)
@@ -647,15 +661,17 @@ bool CCharacterDialogWidget::RenameCharacter()
 }
 
 //*****************************************************************************
-bool CCharacterDialogWidget::RenameVar()
+bool CCharacterDialogWidget::RenameVar(const bool bIsArrayVar)
 //Prompt the user to rename the selected hold variable.
 //
 //Returns: whether a var was renamed
 {
-	if (!this->pVarListBox->ClickedSelection())
+	CListBoxWidget* varListBox = bIsArrayVar ? this->pArrayVarListBox : this->pVarListBox;
+
+	if (!varListBox->ClickedSelection())
 		return false;
 
-	const UINT varID = this->pVarListBox->GetSelectedItem();
+	const UINT varID = varListBox->GetSelectedItem();
 	if (!varID)
 		return false;
 
@@ -674,7 +690,7 @@ bool CCharacterDialogWidget::RenameVar()
 	bool bGoodSyntax;
 	WSTRING wstr;
 	do {
-		wstr = this->pVarListBox->GetSelectedItemText();
+		wstr = varListBox->GetSelectedItemText();
 		const UINT answerTagNo = pEditRoomScreen->ShowTextInputMessage(
 				MID_RenameVariablePrompt, wstr, false, true);
 		if (answerTagNo != TAG_OK)
@@ -686,14 +702,19 @@ bool CCharacterDialogWidget::RenameVar()
 			pEditRoomScreen->ShowOkMessage(MID_VarNameSyntaxError);
 	} while (!bGoodSyntax);
 
+	if (ScriptVars::IsCharacterArrayVar(wstr) != bIsArrayVar) {
+		pEditRoomScreen->ShowOkMessage(L"Can't change var type");
+		return false;
+	}
+
 	if (!pEditRoomScreen->pHold->RenameVar(varID, wstr))
 	{
 		pEditRoomScreen->ShowOkMessage(MID_VarNameDuplicationError);
 		return false;
 	}
 
-	this->pVarListBox->SetSelectedItemText(wstr.c_str());
-	this->pVarListBox->Paint();
+	varListBox->SetSelectedItemText(wstr.c_str());
+	varListBox->Paint();
 	return true;
 }
 
@@ -1827,10 +1848,19 @@ void CCharacterDialogWidget::AddCommandDialog()
 			X_VARTEXTLABEL, Y_VARTEXTLABEL, CX_VARTEXTLABEL, CY_VARTEXTLABEL,
 			F_Small, g_pTheDB->GetMessageText(MID_VarNameText)));
 
+	this->pAddCommandDialog->AddWidget(new CLabelWidget(TAG_ARRAYVAR_TEXTLABEL,
+		X_VARTEXTLABEL, Y_VARTEXTLABEL, CX_VARTEXTLABEL, CY_VARTEXTLABEL,
+		F_Small, L"Variable Name/Expression"));
+
 	this->pVarListBox = new CListBoxWidget(TAG_VARLIST,
 			X_VARLISTBOX, Y_VARLISTBOX, CX_VARLISTBOX, CY_VARLISTBOX, true);
 	this->pVarListBox->AddHotkey(SDLK_RETURN, TAG_OK);
 	this->pAddCommandDialog->AddWidget(this->pVarListBox);
+
+	this->pArrayVarListBox = new CListBoxWidget(TAG_ARRAYVARLIST,
+		X_VARLISTBOX, Y_VARLISTBOX, CX_VARLISTBOX, CY_VARLISTBOX, true);
+	this->pArrayVarListBox->AddHotkey(SDLK_RETURN, TAG_OK);
+	this->pAddCommandDialog->AddWidget(this->pArrayVarListBox);
 
 	CButtonWidget *pVarAddButton = new CButtonWidget(TAG_VARADD, X_VARADD,
 			Y_VARADD, CX_VARADD, CY_VARADD, g_pTheDB->GetMessageText(MID_VarAdd));
@@ -1838,6 +1868,9 @@ void CCharacterDialogWidget::AddCommandDialog()
 
 	CButtonWidget *pVarRemoveButton = new CButtonWidget(TAG_VARREMOVE, X_VARREMOVE,
 			Y_VARREMOVE, CX_VARREMOVE, CY_VARREMOVE, g_pTheDB->GetMessageText(MID_VarRemove));
+	this->pAddCommandDialog->AddWidget(pVarRemoveButton);
+	pVarRemoveButton = new CButtonWidget(TAG_ARRAYVAR_REMOVE, X_VARREMOVE,
+		Y_VARREMOVE, CX_VARREMOVE, CY_VARREMOVE, g_pTheDB->GetMessageText(MID_VarRemove));
 	this->pAddCommandDialog->AddWidget(pVarRemoveButton);
 
 	this->pVarOpListBox = new CListBoxWidget(TAG_VAROPLIST,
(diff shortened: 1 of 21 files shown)


DROD/CharacterDialogWidget.cpp
```diff
@@ -686,14 +702,19 @@ bool CCharacterDialogWidget::RenameVar()
 			pEditRoomScreen->ShowOkMessage(MID_VarNameSyntaxError);
 	} while (!bGoodSyntax);
 
+	if (ScriptVars::IsCharacterArrayVar(wstr) != bIsArrayVar) {
+		pEditRoomScreen->ShowOkMessage(L"Can't change var type");
```
mrimer, line 706: Add MID

DROD/CharacterDialogWidget.cpp
```diff
@@ -1827,18 +1848,30 @@ void CCharacterDialogWidget::AddCommandDialog()
 			X_VARTEXTLABEL, Y_VARTEXTLABEL, CX_VARTEXTLABEL, CY_VARTEXTLABEL,
 			F_Small, g_pTheDB->GetMessageText(MID_VarNameText)));
 
+	this->pAddCommandDialog->AddWidget(new CLabelWidget(TAG_ARRAYVAR_TEXTLABEL,
+		X_VARTEXTLABEL, Y_VARTEXTLABEL, CX_VARTEXTLABEL, CY_VARTEXTLABEL,
+		F_Small, L"Variable Name/Expression"));
```
mrimer, line 1853: Add MID

DROD/CharacterDialogWidget.cpp
```diff
@@ -1883,6 +1928,10 @@ void CCharacterDialogWidget::AddCommandDialog()
 			X_VARVALUELABEL, Y_VARVALUELABEL, CX_VARVALUELABEL, CY_VARVALUELABEL,
 			F_Small, g_pTheDB->GetMessageText(MID_VarOperand)));
 
+	this->pAddCommandDialog->AddWidget(new CLabelWidget(TAG_ARRAYINDEX_LABEL,
+		X_VARVALUELABEL, Y_VARVALUELABEL, CX_VARVALUELABEL, CY_VARVALUELABEL,
+		F_Small, L"Index"));
```
mrimer, line 1933: Add MID

DROD/CharacterDialogWidget.cpp
```diff
@@ -4798,6 +4917,9 @@ void CCharacterDialogWidget::PopulateCommandListBox()
 	this->pActionListBox->AddItem(CCharacterCommand::CC_SetPlayerBehavior, L"Set player behavior");
```
mrimer, line 4917: Add mid for this pre-existing one and the new ones.

DRODLib/Character.cpp
```diff
@@ -5093,6 +5214,57 @@ void CCharacter::SetVariable(const CCharacterCommand& command, CCurrentGame* pGa
...
+		return; //Don't set normal var as an array var
+
+	int arrayIndex = (int)(this->paramF == NO_OVERRIDE ? command.flags : this->paramF);
+	ScriptArrayMap& scriptArrays = pGame->pHold->IsLocalVar(varIndex) ? this->localScriptArrays : pGame->scriptArrays;
+
+	vector<WSTRING> expressions = WCSExplode(command.label, *wszSemicolon);
+	for (vector<WSTRING>::const_iterator expression = expressions.begin();
+		expression != expressions.end(); ++expression) {
+		UINT index = 0;
+		int x = scriptArrays[varIndex][arrayIndex];
```
mrimer, line 5234: We should probably note with a comment that operator[] here will implicitly add a value to the map if the key doesn't already exist.

DRODLib/Character.cpp
```diff
@@ -5093,6 +5214,57 @@ void CCharacter::SetVariable(const CCharacterCommand& command, CCurrentGame* pGa
...
+					x /= operand;
+			break;
+			case ScriptVars::Mod:
+				if (operand)
+					x = x % operand;
+			break;
+			default: break;
+		}
+
+		scriptArrays[varIndex][arrayIndex] = x;
```
mrimer, line 5263: I'm a bit worried about this bloating memory needlessly when the set value is zero. How about if x is 0, then unset the map at arrayIndex to manage memory?

DRODLib/CurrentGame.cpp
```diff
@@ -1994,6 +1997,41 @@ void CCurrentGame::LoadPrep(
...
+		if (!buffer)
+			continue;
+
+		map<int, int> scriptArray;
+		UINT index = 0;
+		UINT size = readBpUINT(buffer, index);
+		while (size) {
+			int key = (int)readBpUINT(buffer, index);
+			int value = (int)readBpUINT(buffer, index);
+			scriptArray[key] = value;
```
mrimer, line 2027: Is it okay to assert(value), i.e., is != 0?

Data/Help/1/scriptarray.html
```diff
@@ -0,0 +1,42 @@
+<html>
+<head>
+    <title>Help for Deadly Rooms of Death</title>
+</head>
+<body bgcolor="#FFFFFF">
+<img src="images/drodlogo.jpg" />
+<hr />
+<h3><u>Using Arrays in Scripts</u></h3>
+    <p>Script variables can be configured as arrays, by starting the variable name with a hash (#) or an at symbol (@).
```
mrimer, line 9: "arrays, by" --> "arrays by"

Data/Help/1/scriptarray.html
```diff
@@ -0,0 +1,42 @@
+<html>
+<head>
+    <title>Help for Deadly Rooms of Death</title>
+</head>
+<body bgcolor="#FFFFFF">
+<img src="images/drodlogo.jpg" />
+<hr />
+<h3><u>Using Arrays in Scripts</u></h3>
+    <p>Script variables can be configured as arrays, by starting the variable name with a hash (#) or an at symbol (@). 
+    Variables starting with a hash a global arrays, while variables starting with an at symbol are local to individual characters.
```
mrimer, line 10: "a global arrays" --> "are global arrays"

Data/Help/1/scriptarray.html
```diff
@@ -0,0 +1,42 @@
...
+<head>
+    <title>Help for Deadly Rooms of Death</title>
+</head>
+<body bgcolor="#FFFFFF">
+<img src="images/drodlogo.jpg" />
+<hr />
+<h3><u>Using Arrays in Scripts</u></h3>
+    <p>Script variables can be configured as arrays, by starting the variable name with a hash (#) or an at symbol (@). 
+    Variables starting with a hash a global arrays, while variables starting with an at symbol are local to individual characters. 
+    Once a variable is created, you cannot change if it is an array variable (but its locality can be changed). Array variables can only store integer values, not text.</p>
```
mrimer, line 11: "change if it is" --> "change whether it is"

Data/Help/1/scriptarray.html
```diff
@@ -0,0 +1,42 @@
...
+    <p>An array variable is effectively a list of values, each associated with a specific index. Therefore, an arbitary number of values can be stored under a single variable name.
```
mrimer, line 12: "a list of values, each associated with a specific index" --> "a set of index-value pairs, where each index has a value (defaulting to 0)"

Data/Help/1/scriptarray.html
```diff
@@ -0,0 +1,42 @@
...
+    Once a variable is created, you cannot change if it is an array variable (but its locality can be changed). Array variables can only store integer values, not text.</p>
+    <p>An array variable is effectively a list of values, each associated with a specific index. Therefore, an arbitary number of values can be stored under a single variable name. 
+    In character scripts, all array indexes are integers, however unlike many programming languages, arrays allow negative indexes to be used.</p>
+
+    <p><u>Setting array values</u></p>
+    <p>Values can be set in an array variable using the <b>Set array var</b> command. The value at a specified index can be set either to an integer value, or an expression that will resolve to such a value.</p>
+    <p>Additionally, multiple array values can be set by providing multiple values or expressions, separated by a semi-colon (;). The first value will be assigned at the specified index, 
+    and seach subsequent value will be assigned to the next index. For example, if the specified index is 3, the first value is assigned to the third array position, the second value to the forth array position, and so on.</p>
```
mrimer, line 18: "seach" -> "each"

Pedantic, but relevant with negative indices:
"if the specified index is 3, the first value is assigned to the third array position, the second value to the forth array position, and so on." -->
"if the specified index is 3, the first value is assigned at position 3, the second value at position 4, and so on."

Data/Help/1/scriptarray.html
```diff
@@ -0,0 +1,42 @@
...
+    <p>The <b>Set array var at</b> command works in the same way, but allows for remote variable assignment, as with <a href="script.html#setvarat">Set var at</a>.</p>
+
+    <p><u>Accessing array values</u></p>
```
mrimer, line 21: Accessing --> Querying

Data/Help/1/scriptarray.html
```diff
@@ -0,0 +1,42 @@
...
+    <p>Like any other script variable, an array value can be used anywhere an integer value is expected. 
+    The one different is that the variable name must be followed by a pair of square brackets ([]) containing an index value. This can be a numeric value, or an another expression that resolves to such a value.
```
mrimer, line 23: different -> difference
"numeric value, or an another" --> "numeric value or an"

Data/Help/1/scriptarray.html
```diff
@@ -0,0 +1,42 @@
...
+    <p>Additionally, multiple array values can be set by providing multiple values or expressions, separated by a semi-colon (;). The first value will be assigned at the specified index, 
+    and seach subsequent value will be assigned to the next index. For example, if the specified index is 3, the first value is assigned to the third array position, the second value to the forth array position, and so on.</p>
+    <p>The <b>Set array var at</b> command works in the same way, but allows for remote variable assignment, as with <a href="script.html#setvarat">Set var at</a>.</p>
+
+    <p><u>Accessing array values</u></p>
+    <p>Like any other script variable, an array value can be used anywhere an integer value is expected. 
+    The one different is that the variable name must be followed by a pair of square brackets ([]) containing an index value. This can be a numeric value, or an another expression that resolves to such a value. 
+    If an index that has not been set is accessed, a value of zero is returned.</p>
```
mrimer, line 24: "a value of zero" -> "a default value of 0"

DRODLib/Character.cpp
```diff
@@ -2992,6 +3062,55 @@ void CCharacter::Process(
...
+					//remote variable set.
+					STOP_COMMAND;
+				}
+
+				bProcessNextCommand = true;
+			}
+			break;
+			case CCharacterCommand::CC_ClearArrayVar:
+			{
+				pGame->scriptArrays[command.x].clear();
```
mrimer, line 3107: I think we need to confirm that command.x is actually referencing a valid map in script Arrays, i.e., via find(). If it's not present, I don't think we should create one implicitly here via operator[]. That might lead to unintended side-effects. Just ignore it if iter == end().

DRODLib/Character.cpp
```diff
@@ -1288,6 +1321,43 @@ int CCharacter::parseFactor(const WCHAR *pwStr, UINT& index, CCurrentGame *pGame
...
+			++index;
+		else
+		{
+			//parse error -- return the current value
+			LogParseError(pwStr, "Parse error (missing close bracket)");
+		}
+
+		UINT varId = pGame->pHold->GetVarID(wVarName.c_str());
+		ScriptArrayMap& scriptArrays = bIsLocalArray ? pNPC->localScriptArrays : pGame->scriptArrays;
+		int val = scriptArrays[varId][arrayIndex];
```
mrimer, line 1359: Here, I think we need to check whether varId is an array var or not.
If not, val should probably default to 0. Otherwise, we're creating a new empty map here, which is wasteful. Only look at pre-existing maps.

I think we should also check whether [arrayIndex] is present in an existing map via find(). If not, return 0. Otherwise, dereference the iterator and return the value. (Don't use operator[arrayIndex] after a find(), or this will wastefully traverse the map structure a second time to locate the key value again!)

DRODLib/Character.cpp
```diff
@@ -1288,6 +1321,43 @@ int CCharacter::parseFactor(const WCHAR *pwStr, UINT& index, CCurrentGame *pGame
 
 		//else: unrecognized identifier -- just return a zero value
 		return val;
+	} else if (pwStr[index] == W_t('#') || pwStr[index] == W_t('@')) {
+		//Find spot where var identifier ends.
```
mrimer, line 1325: Request: add a comment here describing that this logic is parsing an array index, then looking up the value if it is a valid array variable.

DRODLib/Character.cpp
```diff
@@ -5093,6 +5214,57 @@ void CCharacter::SetVariable(const CCharacterCommand& command, CCurrentGame* pGa
...
 }
 
+//*****************************************************************************
+void CCharacter::SetArrayVariable(
+	const CCharacterCommand& command,
+	CCurrentGame* pGame,
+	CCueEvents& CueEvents)
+{
+	const UINT varIndex = command.w;
+	if (pGame->pHold && !pGame->pHold->IsArrayVar(varIndex))
```
mrimer, line 5224: Hmm...I agree that we need to guard against messing with vars that are non-arrays, but where is the logic that turns a non-initialized var into an array in the first place? Wouldn't that be done here?

mrimer (changes requested):

mrimer:
@Hypexion Thanks for the PR! This is cool functionality that would have probably been useful to have a long time ago.

I'm wondering whether I need to be concerned about RAM usage. Considering that game snapshots will store full NPC var copies, what would it take for a player to use up all available RAM and crash the app?

I'm also concerned about address fragmentation in the 32-bit Windows binary in particular. We've had issues with this in the past and have had to be very careful to avoid a lot of random-access map allocs and de-allocs, e.g., in DB acceleration structures. If players go nuts with this functionality, or there's a bad actor, it would be easy to crash the game.

What guards do we want to put in place here? Maybe not allow over a certain high cardinality of entries in a map (i.e., check for map size before a new index assignment)? Even if it's something relatively large like ten thousand, 100k, or a million, I think we need some guardrails.

DRODLib/Character.cpp
Hypexion: The current way things are wired up, script vars being array or non-array is sort of like static typing, based on whether the variable name starts with an array signifier (`#` or `@`). In the same way `CCharacter::SetVariable` checks if a var is local using the pre-built map in `CDbHold`, the check for a var being an array is done using a similar map. This gets constructed when the hold is loaded in, so an array var is effectively always an array, even without being initialized. (the check against changing the type via renaming doesn't use this map because it's not always up to date when editing, but it follows the same principle)

Hypexion:
> I'm wondering whether we need to be concerned about RAM usage. Considering that game snapshots will store full NPC var copies, what would it take for a player to use up all available RAM and crash the app?
> 
> I'm also concerned about address fragmentation in the 32-bit Windows binary in particular. We've had issues with this in the past and have had to be very careful to avoid a lot of random-access map allocs and de-allocs, e.g., in DB acceleration structures. If players go nuts with this functionality, or there's a bad actor, it would be easy to crash the game.
> 
> What guards do we want to put in place here? Maybe not allow over a certain high cardinality of entries in a map (i.e., check for map size before a new index assignment)? Even if it's something relatively large like ten thousand, 100k, or a million, I think we need some guardrails.

I've pushed up some changes to smooth things out, but I definitely see the concerns here. It might be worth restricting arrays to a specific range of indexes. Five or six digit range in positive and negative directions is probably enough for anyone, I would think.

mrimer (approved):

PR #594, opened 2024-06-22
Add script vars to access player weapon and inputs

Hypexion:
New script vars to access some more bits of information:

- `_PlayerWeapon` gives access to the player's base weapon, allowing to be checked separately from the currently wielded weapon, and provides another way to change it, include enabling and disabling the weapon.
- `PlayerLocalWeapon` gives access to the weapon the player is currently wielding, allowing scripted changes of the local weapon without effecting the global weapon, including arming and disarming.
- `_Input` give read-only access to the most recent command the player input.
- `_InputDirection` gives read-only access to the movement direction of the most recent player input, or 4 if a non-movement command was input.

The two input commands required a new member variable in `CCurrentGame`, which is set by `ProcessCommand`. It should not need to be serialized, as its only purpose to allow the access during script processing, which can only happen as part of command processing.

Thread 1: https://forum.caravelgames.com/viewtopic.php?TopicID=46100
Thread 2: https://forum.caravelgames.com/viewtopic.php?TopicID=46142

### DIFF
--- a/Data/Help/1/script.html
+++ b/Data/Help/1/script.html
@@ -709,6 +709,8 @@ You may select these in variable commands to examine or set certain game states.
 	<li><b>_MyScript*</b> - Variables used to inject dynamic values into other scripting commands. <a href="myscript.html">See this section</a> for explanation on how to use them.</li>
 	<li><b>_X</b> and <b>_Y</b> - X and Y position of the Player. Can cause the same potential problems as <b>_MyX</b> and <b>_MyY</b>, so use 'Teleport player to' wheen you need to teleport along two axes.</li>
 	<li><b>_O</b> - Player's current <a href="#valuesorientation">orientation</a>.</li>
+    <li><b>_PlayerWeapon</b> - Can be used to check the player's base weapon, or to indefinitely change their weapon in the same manner as the <a href="#setplayerweapon">Set player weapon</a> command. This value can also be set to -2 to disable the player's weapon, and to -1 to enable it.</li>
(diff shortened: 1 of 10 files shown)


Data/Help/1/script.html
```diff
@@ -709,11 +709,14 @@ <h3><u>Character Script Commands</u></h3>
 	<li><b>_MyScript*</b> - Variables used to inject dynamic values into other scripting commands. <a href="myscript.html">See this section</a> for explanation on how to use them.</li>
 	<li><b>_X</b> and <b>_Y</b> - X and Y position of the Player. Can cause the same potential problems as <b>_MyX</b> and <b>_MyY</b>, so use 'Teleport player to' wheen you need to teleport along two axes.</li>
 	<li><b>_O</b> - Player's current <a href="#valuesorientation">orientation</a>.</li>
+    <li><b>_PlayerWeapon</b> - Can be used to check the player's base weapon, or to indefinitely change their weapon in the same manner as the <a href="#setplayerweapon">Set player weapon</a> command. This value can also be set to -2 to disable the player's weapon, and to -1 to enable it.</li>
+    <li><b>_PlayerLocalWeapon</b> - Can be used to check the weapon currently be used by the player, or to set their current weapon for the duration of the room. This value can also be set to -2 to disarm the player, and to -1 to rearm them, as if touching a Disarm token.</li>
```
mrimer, line 713: currently be used --> currently being used

mrimer (approved):

PR #610, opened 2024-07-23
RPG: Add new monsters to editor + puff effects

kieranmillar:
Adds the two new monsters to the monster editor tab. In order to fit them I had to shrink the serpents from 2 tiles wide to one. It was either those or the rock giant, but reducing the serpents to only the head looked better.

Also adds effects to combat with the puff, wiring up its death effect and replacing blood with fluff particles. The construct still bleeds, that's a change for another day.


mrimer (approved):

mrimer:
Thanks!

PR #611, opened 2024-07-24
New noises for egg spawns and Construct splat

kieranmillar:
Add event for queen spawns which you can now wait for etc.

New egg spawn event plays egg spawning sound.

Add construct death sounds.

Add golem debris effect instead of blood when fighting construct (I thought it had its own effect in classic DROD, but turns out it doesn't)

Has an accompanying PR in the data repo.


mrimer (approved):

mrimer:
Excellent! Thanks!

PR #614, opened 2024-07-25
RPG: Port fix for desync when changing monster appearance

kieranmillar:
https://forum.caravelgames.com/viewtopic.php?TopicID=40625

Game could desync when changing an NPC appearance, this ports over a fix from classic DROD. Looking at the fix, this bug may have been present for changing player roles too.


mrimer (approved):
Thanks!

PR #616, opened 2024-07-30
RPG: Fix characters restoring from wrong room when undoing

kieranmillar:
Second attempt at fixing https://forum.caravelgames.com/viewtopic.php?TopicID=40625

When you undo a room and reload it, the monster list wasn't populating from the explored room, but rather the freshly reloaded room. This caused an issue if a monster had previously used the Set Appearance command, as they have different script commands compared to their logical identity, and so their script commands are only in the explored room monster list.

This creates a variant of the GetCharacterFromScriptId method to pull from the provided explored room.


mrimer (approved):
Thanks!

PR #621, opened 2024-09-01
Remove unfinished equation parser and cache that was not used

EvidentlyCube:
Long long time ago I was working on caching expression parsing to speed up very complex scripting.

It was never finished and the code was just lying there, confusing things needlessly.

This PR is for removal of the code. I am 99% it was not used by anything as for every class I removed I did full source code search and found no references.


mrimer (commented):
Do the VS project files also need to have references to these files removed?

EvidentlyCube:
@mrimer From what I see they don't appear in any of the project files.

PR #623, opened 2024-10-07
More script queries

Hypexion:
A loesely related group of additions to allow more stuff to be queried in scripts.

- `_MonsterSize` script primitive, to check how large a monster at a specific tile is. Works on both heads and pieces.
- `_CleanRooms` script primitive, which can be used to know how many rooms in the level are clean. Parameter can be used to filter the types of rooms checked for (Required/secret/unrequired).
- Added a "Required" option to `Wait for entity`. This looks for required monsters. ([Thread](https://forum.caravelgames.com/viewtopic.php?TopicID=38994))
- `Wait for player state`, which allows checking for invisibility, haste and poweredness.
- `_Combo` predefined var, which accesses the combo number.

(A few unused variable declarations were also trimmed)


mrimer (approved):

mrimer:
Thank you, @Hypexion !

PR #624, opened 2024-10-12
Don't clear spark effects when in process of leaving room

Hypexion:
Removing spark effects in `CGameScreen::ProcessCueEventsBeforeRoomDraw` after `CCurrentGame` has loaded a new room can lead to crashes. To fix this, the spark clearing is skipped if the player has left the room.

Thread: http://forum.caravelgames.com/viewtopic.php?TopicID=46021


schikore (approved):
This is precisely what my local changes looked like, and I can confirm it fixed the problem for me.

mrimer:
Thank you both!

PR #627, opened 2024-10-25
Improvements for key rebinding

Hypexion:
A few improvements to rebindable controls:

- The variable monitor and frame rate commands can now have their inputs configured
- Commands that work in different contexts can share keybinds. Currently the only contexts that matter are playing the game, and editing rooms. This is useful as players coming in from older `5.x` versions will otherwise have clashing bindings.
- Revert the change to some default bindings, as they no longer clash with any other commands.

(Slightly related thread: https://forum.caravelgames.com/viewtopic.php?TopicID=46202)


mrimer (approved):
Thank you!

PR #630, opened 2024-11-10
RPG: Prevent Tendry snoring in the presence of enemy characters

kieranmillar:
Tendry would fall asleep in rooms with custom enemies, which were incorrectly not adding to the danger count of the room.


mrimer (approved):
Thanks!

PR #632, opened 2024-11-12
Fix unset flashing text opacity

kieranmillar:
Fix flashing text color opacity being unset, causing inconsistent results. What value of opacity it was using was essentially random, whatever value happened to be in that part of memory at that time.

This is the same as #631 but for classic DROD instead of RPG.

I believe this is the cause of https://forum.caravelgames.com/viewtopic.php?TopicID=46235


Hypexion:
This probably wants to be pointed at the 5.2 branch

mrimer (approved):

PR #633, opened 2024-11-14
Add more targeting and pathing script hooks

Hypexion:
Adds some more ways for scripts to work with brains and brain-related pathmaps:

- Two `Nearest open room edge` options for `Get natural target`, which perform a horn-like search. One uses the player as the starting position and movement type, the other uses the NPC's.
- `_BrainDist` script primitive, which allows finding brain distance between two arbitary room tiles. ([Thread](https://forum.caravelgames.com/viewtopic.php?TopicID=46178))
- `Wait for brain sensing player` command, which provides an accurate way for a script to determine if monsters in the room will act 'brained'.


DRODLib/CharacterCommand.h
```diff
@@ -200,8 +200,10 @@ namespace ScriptFlag
 		BrainedMonster = 1,
 		BestBrainTile = 2,
 		BestBrainDirection=3,
-		Puff=4,
-		Stalwart=5
+		NearestOpenRoomEdge = 4,
```
mrimer, line 203: Just thought I'd confirm with you that it's okay that we're renumbering these.

DRODLib/CurrentGame.cpp
```diff
@@ -675,6 +675,24 @@ int CCurrentGame::EvalPrimitive(ScriptVars::PrimitiveType ePrimitive, const vect
...
+			int y = params[3];
+			MovementType movement = (MovementType)params[4];
+
+			if (!this->pRoom->IsValidColRow(tX, tY) || !this->pRoom->IsValidColRow(x, y) ||
+				movement >= NumMovementTypes) {
+				return -1;
+			}
+
+			//Make sure pathmap exists
+			this->pRoom->CreatePathMap(tX, tY, movement);
```
mrimer, line 692: Two questions:

Is the intent to override the brained target? If not, wouldn't this approach collide with operations around pre-existing or other pathmaps in the room that manage the same movement type? If they're not intended to collide, wouldn't we need to restore any pre-existing pathmap after performing this calculation?

Also, this approach sounds expensive if it's used more than once per turn!

DRODLib/CharacterCommand.h
Hypexion: Currently the puff and stalwart values aren't being used, so this shouldn't cause a problem. It might be cleaner to remove them, perhaps.

DRODLib/CurrentGame.cpp
Hypexion: Did some more testing it this does weird stuff if it's called before monsters making brained moves. So a rethink is definitely in order. We might paths for this held somewhere else, which can also allow caching to reduce the cost of doing this multiple times.

Hypexion: I think this problem technically also exists with `Nearest open edge to me`, but since edge-finding uses the `FORCE` movement types, it doesn't affect monster pathing. Then when a horn is used that sets the target so it doesn't cause a problem there.

mrimer: Thanks for checking on this. How about we use dedicated movement slots in the room data object to store these ad-hoc scripted pathmaps, say, "EPHEMERAL + x" movement index, so they won't clobber the base types.

Hypexion: Tried a few possible solutions for this. Ultimately, the smoothest one was to add a new set of pathmaps in `CDbRoom` that I've called `pExtraPathMap`. For the moment it's only used by the `_BrainDist` logic, but could be used elsewhere if needed.

mrimer (approved):

PR #636, opened 2024-11-17
RPG: Add Minimap treasure behavior

Hypexion:
This is effectively a flag you can add to characters so that they make the minimap yellow. To reduce the possibility of mischief, it only works for characters that are visible in some way and haven't ended.


mrimer (approved):
Thanks!

PR #638, opened 2024-11-17
Update Door splash image positioning.

mrimer:
Update RPG version to "2.0.0.#-dev"


mrimer:
Please use the new Door.png with this commit.

PR #644, opened 2024-11-23
RPG: Prevent Assert failure when firetrap turns itself off

kieranmillar:
Prevent an assert failure when a firetrap turns itself off by moving the processing of destroying mist before it processes blowing up bombs and kegs. This caused it to turn into an off firetrap before checking if it was allowed to destroy mist.


mrimer (approved):

mrimer:
Thanks!

PR #646, opened 2024-11-24
RPG: Only allow attacking when back is turned if movement is not blocked

kieranmillar:
"Attack when adjacent" behaviour cannot attack when movement is blocked, but the same was not also true of "attack when back is turned" behaviour. This copies across the same check to "attack when back is turned".

https://forum.caravelgames.com/viewtopic.php?TopicID=46269


mrimer (approved):
Thanks!

PR #647, opened 2024-11-24
RPG: Let explosions destroy equipment

kieranmillar:
Allows equipment to be destroyed by explosions. This also destroys empty slots as they technically count as equipment.

https://forum.caravelgames.com/viewtopic.php?TopicID=46275


kieranmillar:
I'm closing this change as upon reflection this change has too many potential consequences.

PR #653, opened 2024-11-30
RPG: Additional script commands from TSS 5.2

Hypexion:
Brings over some more waiting commands from mainline 5.2, although due to differences between the two games, they are some functional differences.

- `Reset _MyScript variables`, which is identical to how it works in DROD
- `Wait for weapon`, which does not allow for looking for a specific weapon, due to how weapons work in RPG.
- `Wait for open tile` has fewer options in the ignore flags, as RPG has less wait flags.
- `Wait for item group` and `Wait while item group` have some different groups, as RPG has different elements. Otherwise the logic is the same.


mrimer (approved):
Thank you for adding this!

PR #655, opened 2024-12-01
RPG: Small UI improvements

Hypexion:
- Adds a pop-out menu for health potions in the editor
- Empty equipment slots aren't described, fixing an issue where they would get a tooltip saying they were metal
- Holding control when right-clicking suppresses the tooltip, only showing highlighting
- "Surprise from behind" is now included in monster descriptions when appropriate


mrimer (approved):
Thank you!

PR #656, opened 2024-12-01
RPG: Fix crash with Attack in front behavior

Hypexion:
When a character with "Attack in front" attacks, it now checks to see if its base type is an eye, and if so, sends an "Evil eye woke" event. However, if the character doesn't have a `pCustomChar` set, the game crashes.

The logic has been made more robust to prevent this problem.


mrimer (approved):
Thanks!

PR #657, opened 2024-12-02
RPG: Add six additional function primitives

Hypexion:
Ports over the following function primitives that exist in TSS to RPG:

- `_facing` - The orientation that is closest to the given (x,y) vector
- `_rotateDist` - How many rotations are between (o1,o2)
- `_ArrowDir` - Which way is an arrow at (x,y) pointing
- `_RoomTile` - The tile type at (x,y,layer)
- `_MonsterType` - Type of monster at (x,y)
- `_CharacterType` - Type of NPC at (x,y)


mrimer (approved):
Thank you!

PR #658, opened 2024-12-07
RPG: Add more room interaction behaviours

Hypexion:
More behaviours that change how stuff interacts with thigns in the room:

- `Cut tarstuff anywhere`, which is effectively the spear behaviour, allowing any tarstuff to be damaged on any tile.
- `Wall and mirror safe` which stops the player from breaking things
- `Heatproof` and `Fireproof` which protect from hot tiles and firetraps respectively
- `Mist protection` which negates mist's DEF nullifying properties


mrimer (approved):
Thank you so much!

PR #662, opened 2024-12-08
RPG: Change default shovel score value

kieranmillar:
Change the default score value for shovels to 1 per shovel, so players that aren't messing with score multipliers get some value from them. This will cause it to show up by default in all the score checkpoint breakdowns, including for older holds, but so do other commonly unused resources like Skeleton keys, so I figure it's fine.


mrimer (approved):
Thanks! Yeah, this sounds fine for a 2.0 release.

PR #666, opened 2024-12-09
RPG: Fix disabled equipment retaining combat abilities

Hypexion:
If a piece of equipment is disabled, it's behaviours that affect combat shouldn't apply.

I'm fairly certain this change won't break anything. In addition to require very specific circumstances to manifest, I think most of these behaviours are used in published holds, in part because people are unaware of what they do.

Thread: http://forum.caravelgames.com/viewtopic.php?TopicID=46292


mrimer (commented):
Approved

mrimer:
Merging to try this out, as per the forum topic.

PR #674, opened 2024-12-16
RPG: Rename "aspect" to "type" for custom weaknesses

kieranmillar:
The word aspect for this is confusing, type would be much better understood.


Hypexion:
It looks like you missed the entries in `Speech.uni`

PR #682, opened 2024-12-17
RPG: Trigger post-combat growths when fighting mothers

kieranmillar:
Fighting mothers bypasses the usual monster defeat code because we often don't want to kill them, however the expansion of post-combat growing elements lived there. This adds those growths to the special case carved out for fighting mothers.

Forum thread: https://forum.caravelgames.com/viewboard.php?BoardID=277

While I was there, found an edge case where clicking on a bomb in the editor could cause a crash if it reaches any tarstuff because there's no CurrentGame instance when simulating a bomb explosion.


mrimer (approved):
Thanks for fixing this!

PR #695, opened 2024-12-23
RPG: Add wall-dwelling behavior for NPCs

Hypexion:
Adds the Wall-dwelling behaviour, that makes NPCs die if they aren't on a solid tile, just like seep. To make this work, a few functions have been shuffled around.

As a related change, all monsters with water movement are now labled as "water-dwelling", not just waterskippers.

Thread: https://forum.caravelgames.com/viewtopic.php?TopicID=46193


mrimer (approved):
Thanks!

PR #697, opened 2024-12-23
RPG: Prevent flickering of characters and rock giants with cloud shadows

kieranmillar:
Characters and rock giants would flicker when cloud shadows were enabled due to their tiles not correctly being marked dirty.

Forum thread: https://forum.caravelgames.com/viewtopic.php?TopicID=45390


mrimer (approved):
Thanks!

PR #698, opened 2024-12-24
Fix bounds checking for some Wait for commands

Hypexion:
Bounds checking isn't being done properly for some script commands, leading to asserts if out-of-bounds areas are provided via _MyScript injection. How this is handled was already handled inconsistently between different commands, so a few different fixes have been done.

Thread: http://forum.caravelgames.com/viewtopic.php?TopicID=46324


mrimer (approved):
Thanks!

PR #702, opened 2024-12-31
RPG: Persist when a script has ended

kieranmillar:
Undo PR #694 and replace with a proper solution. 694 accidentally blew up linking monsters with a NONE identity which had all manner of bad side effects.

This is a another attempt at fixing https://forum.caravelgames.com/viewtopic.php?TopicID=46010

When a character ends their script, there's a brief period of time until you leave the room where the character still exists but the script is done. When you preview another room before leaving the current one, a temporary game is created, but the fact the character's script had ended was not persisted. This caused the room preview to re-process that character's script from the beginning as part of turn 0 processing, which in this case included an Appear command.


mrimer (approved):

mrimer:
Thanks!

PR #704, opened 2025-01-07
RPG: Properly initialise and plot new OLayer elements

kieranmillar:
Certain actions, such as reflecting the room would cause assertions with some new O Layer elements. This is because we had overlooked properly initializing them and updating the values when calling Plot.


mrimer (approved):
Thanks!

PR #706, opened 2025-01-09
RPG: Prevent some script commands processing when room previewing

kieranmillar:
Forum thread: https://forum.caravelgames.com/viewtopic.php?TopicID=46392

When room previewing, we need to do a turn 0 process of the scripts in the room, so that e.g. custom monsters will have stats so we can look at them.

Some commands should not work in this case, such as checking if the player is touching an NPC, because they should not be in the room. This PR blocks some important commands form processing in room preview mode. Most importantly are score checkpoints and autosaves. These were submitting (invalid) scores to the spider, and with autosaves you could actually warp to other rooms in the game by loading the newly-created save.

### DIFF
--- a/drodrpg/DRODLib/Character.cpp
+++ b/drodrpg/DRODLib/Character.cpp
@@ -2361,27 +2361,31 @@ void CCharacter::Process(
 			break;
 			case CCharacterCommand::CC_AddRoomToMap:
 			{
-				if (bRoomBeingDisplayedOnly) break;
-				//Add room at (x,y) to player's mapped rooms.
-				const UINT roomID = pGame->pLevel->GetRoomIDAtCoords(
-						command.x, pGame->pLevel->dwLevelID*100 + command.y);
-				if (roomID)
+				if (!bRoomBeingDisplayedOnly)
 				{
-					const bool bMarkExplored = command.w != 0;
-					pGame->AddRoomToMap(roomID, bMarkExplored);
-					CueEvents.Add(CID_LevelMap, new CAttachableWrapper<UINT>(
+					//Add room at (x,y) to player's mapped rooms.
+					const UINT roomID = pGame->pLevel->GetRoomIDAtCoords(
+						command.x, pGame->pLevel->dwLevelID * 100 + command.y);
+					if (roomID)
+					{
+						const bool bMarkExplored = command.w != 0;
+						pGame->AddRoomToMap(roomID, bMarkExplored);
+						CueEvents.Add(CID_LevelMap, new CAttachableWrapper<UINT>(
 							bMarkExplored ? T_MAP_DETAIL : T_MAP));
+					}
 				}
 				bProcessNextCommand = true;
 			}
 			break;
 			case CCharacterCommand::CC_Autosave:
 			{
-				if (bRoomBeingDisplayedOnly) break;
-				//Autosave with identifier 'label'.
-				WSTRING saveName = pGame->ExpandText(command.label.c_str(), this);
-				if (pGame->Autosave(saveName))
-					CueEvents.Add(CID_Autosave);
+				if (!bRoomBeingDisplayedOnly)
+				{
+					//Autosave with identifier 'label'.
+					WSTRING saveName = pGame->ExpandText(command.label.c_str(), this);
+					if (pGame->Autosave(saveName))
+						CueEvents.Add(CID_Autosave);
+				}
 				bProcessNextCommand = true;
 			}
 			break;
@@ -3120,19 +3124,20 @@ void CCharacter::Process(
 			break;
 			case CCharacterCommand::CC_ScoreCheckpoint:
 			{
-				if (bRoomBeingDisplayedOnly) break;
-				//Defines a scoring point with identifier 'label'.
-				const bool bNotFrozen = !this->pCurrentGame->Commands.IsFrozen();
-				if (bNotFrozen ||  //when playing back commands, don't do this stuff
-						this->pCurrentGame->IsValidatingPlayback()) //unless we're validating
+				if (!bRoomBeingDisplayedOnly)
 				{
-					CDbMessageText *pScoreIDText = new CDbMessageText();
-					*pScoreIDText = command.label.c_str();
-					CueEvents.Add(CID_ScoreCheckpoint, pScoreIDText, true);
-					if (bNotFrozen)
-						const_cast<CCurrentGame*>(this->pCurrentGame)->WriteScoreCheckpointSave(command.label);
+					//Defines a scoring point with identifier 'label'.
+					const bool bNotFrozen = !this->pCurrentGame->Commands.IsFrozen();
+					if (bNotFrozen ||  //when playing back commands, don't do this stuff
+						this->pCurrentGame->IsValidatingPlayback()) //unless we're validating
+					{
+						CDbMessageText* pScoreIDText = new CDbMessageText();
+						*pScoreIDText = command.label.c_str();
+						CueEvents.Add(CID_ScoreCheckpoint, pScoreIDText, true);
+						if (bNotFrozen)
+							const_cast<CCurrentGame*>(this->pCurrentGame)->WriteScoreCheckpointSave(command.label);
+					}
 				}
-
 				bProcessNextCommand = true;
 			}
 			break;
@@ -3350,16 +3355,19 @@ void CCharacter::Process(
 
 			case CCharacterCommand::CC_SetPlayerAppearance:
 			{
-				if (bRoomBeingDisplayedOnly) break;
-				if (this->bIfBlock)
+				if (!bRoomBeingDisplayedOnly)
 				{
-					//As an If condition, this acts as a query that is true when
-					//the player is in this role.
-					if (player.wIdentity != command.x)
-						STOP_COMMAND;
-				} else {
-					//Sets player's identity to entity X.
-					pGame->SetPlayerRole(command.x);
+					if (this->bIfBlock)
+					{
+						//As an If condition, this acts as a query that is true when
+						//the player is in this role.
+						if (player.wIdentity != command.x)
+							STOP_COMMAND;
+					}
+					else {
+						//Sets player's identity to entity X.
+						pGame->SetPlayerRole(command.x);
+					}
 				}
 				bProcessNextCommand = true;
 			}
--- a/drodrpg/DRODLib/Character.cpp
+++ b/drodrpg/DRODLib/Character.cpp
@@ -1807,6 +1807,9 @@ void CCharacter::Process(
 	CDbRoom& room = *(pGame->pRoom);
 	CSwordsman& player = *pGame->pPlayer;
 
+	// Some commands should not be executed when previewing a room remotely
+	const bool bRoomBeingDisplayedOnly = pGame->IsRoomBeingDisplayedOnly();
+
 	//Keep track of swordsman's orientation on previous and current turn.
 	this->wLastSO = this->wSO;
 	this->wSO = player.wO;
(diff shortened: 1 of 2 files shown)


drodrpg/DRODLib/Character.cpp
```diff
@@ -2358,6 +2361,7 @@ void CCharacter::Process(
 			break;
 			case CCharacterCommand::CC_AddRoomToMap:
 			{
+				if (bRoomBeingDisplayedOnly) return;
```
mrimer, line 2364: I'm wondering whether it makes sense to "break" here to skip this command and let other safe commands run in the command sequence, instead of simply returning/halting immediately when one of these "unsafe" commands is encountered. Thoughts?

kieranmillar: I agree, it is much better to allow commands that would not block processing to only be skipped over rather than stopping further processing of that script. I have reviewed the commands and changed the ones I think it is best to break on instead of return.

drodrpg/DRODLib/Character.cpp
```diff
@@ -2358,6 +2361,7 @@ void CCharacter::Process(
 			break;
 			case CCharacterCommand::CC_AddRoomToMap:
 			{
+				if (bRoomBeingDisplayedOnly) break;
```
mrimer, line 2364: Thanks for making this change.
Recall lower down in the case's code block that `bProcessNextCommand = true;` needs to be executed before breaking in order to process the next commands. How about doing this instead:
`if (!bRoomBeingDisplayedOnly) { <do all of the code before that line> }`

drodrpg/DRODLib/Character.cpp
```diff
@@ -2490,6 +2497,7 @@ void CCharacter::Process(
 			break;
 			case CCharacterCommand::CC_WaitForPlayerToTouchMe:
 			{
+				if (bRoomBeingDisplayedOnly) return;
```
mrimer, line 2500: I'm thinking that instead of a `return` here, we should probably execute `STOP_COMMAND` instead. This is so that if this command is being used as part of an "if" condition, it will return false and continue script execution in the else block, etc.

drodrpg/DRODLib/Character.cpp
```diff
@@ -3200,6 +3209,7 @@ void CCharacter::Process(
 			break;
 
 			case CCharacterCommand::CC_LevelEntrance:
+				if (bRoomBeingDisplayedOnly) return;
```
mrimer, line 3217: It is useful to place the `return` on a new line (indented, of course), so that when running in the debugger, one can place a breakpoint on the `return` to easily catch when it happens.

drodrpg/DRODLib/Character.cpp
```diff
@@ -3340,6 +3350,7 @@ void CCharacter::Process(
 
 			case CCharacterCommand::CC_SetPlayerAppearance:
 			{
+				if (bRoomBeingDisplayedOnly) break;
```
mrimer: I'm wondering whether this check is better placed only in the else block, because it seems like it could work fine for an "if" condition check without harmful side-effects.

drodrpg/DRODLib/Character.cpp
```diff
@@ -2481,6 +2487,7 @@ void CCharacter::Process(
 			break;
 			case CCharacterCommand::CC_WaitForPlayerToMove:
 			{
+				if (bRoomBeingDisplayedOnly) return;
```
mrimer, line 2494: Hmm...I'm wondering how to handle this one when it an "if" condition. I don't think we need to break here. Does this seem right instead:
`if (bRoomBeingDisplayedOnly || !DidPlayerMove(command, player, nLastCommand))
 STOP_COMMAND;`

drodrpg/DRODLib/Character.cpp
kieranmillar: I didn't realize I had to set that value to true, let me amend everything accordingly.

drodrpg/DRODLib/Character.cpp
```diff
@@ -1807,6 +1807,9 @@ void CCharacter::Process(
 	CDbRoom& room = *(pGame->pRoom);
 	CSwordsman& player = *pGame->pPlayer;
 
+	// Some commands should not be executed when previewing a room remotely
```
mrimer, line 1810: Let's update the comment to state
`// Some commands should not have side effects when previewing a room remotely`

drodrpg/DRODLib/Character.cpp
kieranmillar: Good shout on checking for ifs. However in the case of wait for player to touch me, and many of the other commands that do things with the player, they should really all fail all the time if you're room previewing, one of the major issues that lead to this bug is that the player maintains his X and Y coordinates when room previewing so many characters were now thinking the player was touching them when the player wasn't even in the room.

If we're checking a condition of "is the player doing X" and we're not even looking at the room the player is in, I think the answer should always be "no".

drodrpg/DRODLib/Character.cpp
kieranmillar: I think that will be fine, although in general I don't know if I want checks on player state to work if the player isn't even in the room, but in this particular case I think this will be OK.

drodrpg/DRODLib/Character.cpp
mrimer: That's right. So, to get the script flow to process cleanly in all cases, we want to do this:
`player.wX == this->wX && player.wY == this->wY && !bRoomBeingDisplayedOnly) ...`

drodrpg/DRODLib/Character.cpp
```diff
@@ -3340,15 +3355,19 @@ void CCharacter::Process(
 
 			case CCharacterCommand::CC_SetPlayerAppearance:
 			{
-				if (this->bIfBlock)
+				if (!bRoomBeingDisplayedOnly)
```
mrimer: My new comment here might have been lost. To restate:
I think we want to perform this check only in the else block, where there might be a side effect, i.e.,
`if (!bRoomBeingDisplayedOnly) pGame->SetPlayerRole(command.x);`

drodrpg/DRODLib/Character.cpp
kieranmillar: Sure, seems fine

drodrpg/DRODLib/Character.cpp
kieranmillar: I left this one as a return, but put it on a new line. We never want to label jump if this one fails.

drodrpg/DRODLib/Character.cpp
kieranmillar: Yeah sorry there is some confusion as you have been making comments while I've been working on this so there's pushes happening while you're reviewing. I'm going to go make dinner to give you time to finish reviewing before pushing any more changes.

drodrpg/DRODLib/Character.cpp
```diff
@@ -2482,14 +2491,17 @@ void CCharacter::Process(
...
+				if (bRoomBeingDisplayedOnly || !DidPlayerMove(command, player, nLastCommand))
 					STOP_COMMAND;
 
 				bProcessNextCommand = true;
 			}
 			break;
 			case CCharacterCommand::CC_WaitForPlayerToTouchMe:
 			{
+				if (bRoomBeingDisplayedOnly)
+					STOP_COMMAND;
```
mrimer, line 2503: This doesn't work quite right -- the rest of the block will still execute. We need to do this:
`if (player.wX == this->wX && player.wY == this->wY && !bRoomBeingDisplayedOnly) ...`

mrimer (approved):
I think everything will work as-is and we can merge if you're good with these changes. Just let me know if you think we're ready to merge -- though one thing that will be nice is to squash all the commits into one, since we don't need to keep them separate. I'll do that when you confirm we're ready to merge.

Hypexion:
Potential concern from me: Since you're putting the check in the `Process` block, I think the conditional commands will still evaluate as if the player is in the room when using them in the logic block commands.

mrimer:
> Potential concern from me: Since you're putting the check in the `Process` function, I think the conditional commands will still evaluate as if the player is in the room when using them in the logic block commands.

That's right...or rather, the commands will still evaluate with the player not being found in the room by any of the checks.

Would you describe a case of how this could be a problem?

Hypexion:
> > Potential concern from me: Since you're putting the check in the `Process` function, I think the conditional commands will still evaluate as if the player is in the room when using them in the logic block commands.
> 
> That's right...or rather, the commands will still evaluate with the player not being found in the room by any of the checks.
> 
> Would you describe a case of how this could be a problem?

It's the original issue, but happening in another place. To make `Wait for Any` and the other logic block commands work without duplicating a bunch of code, I refactored various checks out of the big switch in `Process` and moved them to separate functions. There's another function, `EvaluateConditionalCommand`, that also uses those functions. The room preview check needs to be either there as well, or in the actual evaluation functions, or the commands that query the player's state will still be able to pass.

mrimer:
> It's the original issue, but happening in another place. To make `Wait for Any` and the other logic block commands work without duplicating a bunch of code, I refactored various checks out of the big switch in `Process` and moved them to separate functions. There's another function, `EvaluateConditionalCommand`, that also uses those functions. The room preview check needs to be either there as well, or in the actual evaluation functions, or the commands that query the player's state will still be able to pass when used in a logic block.

Thanks for the elaboration. Sorry, I don't think I'm following. To restate my understanding -- if the original issue is that the player can sometimes be observed by scripts in a room preview, and we're adding checks to not detect the player, then how is the original issue remaining?

kieranmillar:
I see what hyperme is saying. When using a command as the condition for an if, there's now another way of doing that, found in the method EvaluateConditionalCommand further down in Character.cpp.

To avoid duplication, much of the logic inside each case statement inside Process() was moved out into its a separate method so both Process and EvaluateConditionalCommand could call that to do the part of the code where they check if it works as part of an If.

I have only added my checks to Process, overlooking this other part of the code. I'll amend so that the shared area has check to see if the room is being displayed where applicable.

Hypexion:
> > It's the original issue, but happening in another place. To make `Wait for Any` and the other logic block commands work without duplicating a bunch of code, I refactored various checks out of the big switch in `Process` and moved them to separate functions. There's another function, `EvaluateConditionalCommand`, that also uses those functions. The room preview check needs to be either there as well, or in the actual evaluation functions, or the commands that query the player's state will still be able to pass when used in a logic block.
> 
> Thanks for the elaboration. Sorry, I don't think I'm following. To restate my understanding -- if the original issue is that the player can sometimes be observed by scripts in a room preview, and we're adding checks to not detect the player, then how is the original issue remaining?

It's to do with how the various logic block commands are processed. Rather than setting a flag to say a logic block is being processed, then continuing with normal processing, they instead evaluate all the conditions between their start and end commands as part of processing the start command. e.g. When `CC_LogicalWaitAnd` is processed, that calls `CCharacter::EvaluateLogicalAnd`, which does its own evaluation of conditional commands until one resolves to false, or it reachs a `CC_LogicalWaitEnd` command. Since this is done outside of the main processing command, anything in there doesn't effect checks happening in logic blocks.

Kieran has now moved the checks into the shared functions, which plugs this hole.

kieranmillar:
I'm ready to merge now if everyone else is, and yes please do a squash.

mrimer (approved):
Ah, I see now. The additional changes in the latest commit make sense to me.

@Hypexion , it sounds like you're good with me merging this PR now?

Hypexion:
> Ah, I see now. The additional changes in the latest commit make sense to me.
> 
> @Hypexion , it sounds like you're good with me merging this PR now?

This all looks good to me now.

PR #707, opened 2025-01-10
RPG: Add Tile Lights to persisted room data

kieranmillar:
Forum thread: https://forum.caravelgames.com/viewtopic.php?TopicID=46204

The ExploredRoom data doesn't store everything to keep storage space down, for example it does not store orb connections. Another thing it didn't store was TileLights, but unfortunately those can change, wall lights can be turned off with the Activate Item At script command. This change adds TileLights to the ExploredRoom so it can be persisted.

This is my first time trying to change the database structure so there may be a bunch of gotchas I'm missing. This code seems to load old saves etc. just fine, if the TileLights data is empty in Explored Rooms it will simply use the data from the normal room data instead.

For whatever reason, my local copy of the game tends to crash on some weird stuff, for example I can't fast travel West without that always crashing when working out the path. The Explored Room data was another instance, it really does not like ever trying to initialize a c4_Bytes via copying another instance that has zero content. You will see I changed that code to initialize SquaresBytes as empty and then replace it if the copy has non-zero size. This is possible if the room is on the map but never been explored (e.g. visited in a previous playthrough). This was simply crashing my game every time I tried to preview a room in a level where that was the case. I would have much preferred to do this check inside the c4_Bytes constructor instead of everywhere we use the constructor, but it's a part of the metakit dependency so I can't change it.


mrimer (approved):
This LGTM. @Hypexion , would you please also review and let us know if you're okay with me merging this PR now?

Hypexion:
If it's working cleanly, I don't have objections. Being able to bundle db changes in with the features that require them is preferable if that's possible.

kieranmillar:
I believe this is good to be merged? (If it wasn't clear, this is fully feature complete, saving and loading the data and replacing the room data with it if it's present is the full extent of the use case)

PR #711, opened 2025-01-15
Add ability to toggle cloud sync activation

Hypexion:
The people want to disable cloud sync sometimes. This adds a button to do that, and makes it so that you can reactivate the sync afterwards without having to delete your data.

Thread: https://forum.caravelgames.com/viewtopic.php?TopicID=42719

### DIFF
--- a/DROD/SettingsScreen.cpp
+++ b/DROD/SettingsScreen.cpp
@@ -57,6 +57,12 @@
 
 using namespace InputCommands;
 
+//Static values for storing cloud profile information
+namespace {
+	static WSTRING s_cloudOriginalName;
+	static time_t s_cloudCreated;
+};
+
 //Tabbed menu info.
 #define PERSONAL_TAB (0)
 #define GAS_TAB (1)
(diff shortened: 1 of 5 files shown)


Hypexion:
Currently a draft PR in case there's something that should get done as part of reactivating to make sure things are synced correctly.

Hypexion:
> This looks good to me so far.
> 
> I'm not clear on where the deactivation check is performed in order to avoid executing the cloud sync process itself. Would you please point that out to me?

At the moment what's happening is that the `CloudActivated` setting gets flipped back to false when you deactivate. Since it's false, none of the cloud stuff that checks for it being true will happen.

DROD/SettingsScreen.cpp
```diff
@@ -1488,6 +1515,22 @@ void CSettingsScreen::SetWidgetStates()
...
+//************************************************************************************
+void CSettingsScreen::SetCloudWidgetStates()
+//Enable/disable widgets depending on state. Separate from SetWidgetStates due to how
+//cloud settings are handled.
+{
+	CDbPlayer* pPlayer = g_pTheDB->GetCurrentPlayer();
+
+	CButtonWidget* pButton = DYN_CAST(CButtonWidget*, CWidget*,
+		GetWidget(TAG_CLOUD_DEACTIVATE));
+	pButton->Enable(pPlayer->Settings.GetVar(Settings::CloudActivated, false));
```
mrimer, line 1531: Should this setting be flipped? i.e., when CloudActivated is true, the deactivated button should be set to false, and v.v.

Hypexion: The deactivate button is the button that deactivates the cloud sync, so we want it to be enabled when cloud sync is on, and off otherwise.
Maybe we want the activate button to enable/disable based on the profile setting as well.

mrimer: Ohhh...I get it now. I was thinking this was a check box (toggle) button, not a regular button.

How about instead of an "activate" and "deactivate" button, we just use one check box to drive both actions? I believe they are mutually exclusive, correct?

Hypexion: I'll have a look about that, although I'm not sure if it will mesh with the use of dialog boxes.

mrimer: Good point. I'm wondering about this too. I'm pretty sure that when the box changes state, we can pop up the dialog box as before, and if the player doesn't confirm the action, we can revert the checkbox state.

PR #712, opened 2025-01-17
Include count of challenges on challenges dialog

Hypexion:
Thread: https://forum.caravelgames.com/viewtopic.php?TopicID=46412


mrimer (approved):
Thank you!

PR #720, opened 2025-01-23
RPG: Ensure InitialIdentity is correct when changing character type in editor

kieranmillar:
There is some odd lingering effect hiding somewhere in the editor where, when you copy a character then change its type, the InitialIdentity value is still the same as the old character type, causing it to run the wrong character script. Most of the time it is unset, so not sure in which circumstance it was being set.

Instead, we now always set the initial identity when you change the type, ensuring it's always correct.

P.S. Due to dumb things on my part, my RPG 1.3 branch has some lingering merge commits that made it into this PR, please squash when merging this PR.


mrimer (approved):

mrimer:
Thanks, @kieranmillar . Sorry, would you please rebase this change again on the latest commit version?

kieranmillar:
There, that should be rebased now. Please squash when merging.

mrimer:
Thanks, Kieran. I'm now seeing the overhead tile commits included in your PR. I'm not sure why they're there. I've had this kind of rebasing issue before, and it can be confusing. I'm thinking it might be cleanest to abandon this dev branch and start the process fresh again with your (1-line?) change placed on a brand new branch created at the current tip of the main branch.

PR #726, opened 2025-02-12
Fix position resetting when editing default script

Hypexion:
Thread: https://forum.caravelgames.com/viewtopic.php?TopicID=46463


mrimer (approved):
Thanks!

PR #727, opened 2025-02-15
Add some helpful visual effects

Hypexion:
Adds some visual effects for certain situations:

1. When a platform move is blocked due a heavy object, a bump effect is now applied to all heavy objects on that platform (or raft). Thread: http://forum.caravelgames.com/viewtopic.php?TopicID=46474
2. Bumping door when Halph is in the room now has visual effects. If Halph can't open the door, a cross appears on the bumped tile and Halph. It can be opened, a dotted line appears between Halph and the target orb or plate. The dotted line is also shown when bump a door Halph is already opening. Thread: http://forum.caravelgames.com/viewtopic.php?TopicID=46231
3. If Halph's path becomes blocked, a cross appears on Halph and his target.
4. The new dotted line effect is available for use in scripts.


mrimer (approved):

mrimer:
Thank you!

PR #731, opened 2025-02-19
RPG: Move to 2.0 data files

Hypexion:
Updates all the constants used to determine which files and directories to use to be `2.0` or `2_0`. The `drodrpg1_0.dat` is now `drodrpg2_0.dat`, and has been updated to contain all existing message strings.

Additionally, a few messages missing from the `.uni` have been added, along with a little bit of rearrangement. The messages haven't been removed from `CDbBase` yet, since a fully functional `dat` file for 2.0 isn't publically available.

Furthermore, the suggested change to how explored room state has been made. 1.0 -> 2.0 is a good place to do this, as it would be annoying to migrate existing data files for this change.


drodrpg/DRODLib/CurrentGame.cpp
```diff
@@ -4936,7 +4939,7 @@ void CCurrentGame::AddRoomsPreviouslyExploredByPlayerToMap(
 
 	for (CIDSet::const_iterator it=this->PreviouslyExploredRooms.begin(); it!=this->PreviouslyExploredRooms.end(); ++it)
 	{
-		AddRoomToMap(*it, bMakeRoomsVisible, false);
+		AddRoomToMap(*it, MapState::Preview, false);
```
mrimer, line 4942: How is bMakeRoomsVisible used now?

drodrpg/DRODLib/CurrentGame.cpp
```diff
@@ -383,26 +383,29 @@ void CCurrentGame::AddRoomToMap(
...
 	} else {
 		pExpRoom = new ExploredRoom();
 		this->ExploredRooms.push_back(pExpRoom);
 		pExpRoom->roomID = roomID;
-		pExpRoom->bMapOnly = true;
+		pExpRoom->mapState = mapState;
 		pExpRoom->bSave = bSaveRoom;
 	}
 
 	//Room marked only on the map becomes fully visible (as if explored) if bMarkRoomVisible is set.
```
mrimer, line 406: Update comment

drodrpg/DRODLib/CurrentGame.cpp
Hypexion: It's not used. Since the function is never used without the default arguments, I've removed bMakeRoomsVisible.

mrimer (approved):

mrimer:
Thanks, @Hypexion !

PR #732, opened 2025-02-19
RPG: Fix disabled equipment behaviours still applying

Hypexion:
For master branch to allow a 1.2 version to be released with this fix.

Thread: https://forum.caravelgames.com/viewtopic.php?TopicID=46292


mrimer (approved):
Thanks, @Hypexion!

PR #737, opened 2025-02-22
Fix out of bounds error in BuildUtil::BuildTilesAt

Hypexion:
An oversight with tile numbers meant a bad index could be used for the `TILE_LAYER` array. Some platforms are polite enough to crash instead of allow more worrisome possibilities. This is the root cause of some reported crashes.

Thread 1: https://forum.caravelgames.com/viewtopic.php?TopicID=46270
Thread 2: https://forum.caravelgames.com/viewtopic.php?TopicID=46273


mrimer (approved):
Thanks!

PR #748, opened 2025-03-04
RPG: Fix position of MapState in propTypeStr order

Hypexion:


drodrpg/DRODLib/DbProps.h
```diff
@@ -637,7 +637,7 @@ extern const char propTypeStr[P_Count][26]
 	"IsSecret",
 	"LanguageCode", "LastUpdated", "Left",
 	"LevelID",
-	"LitFuses", "MapOnly", "MapState", "MapMarker",
+	"LitFuses", "MapOnly", "MapMarker", "MapState",
```
mrimer, line 640: Thank you for catching this!
FYI - these strings are intended to be listed in alphabetical/lexicographical order. Should "MapMarker" actually be moved before "MapOnly" everywhere this is relevant?

Hypexion: Chasing down the ordering doesn't seem worth it, except for in `propTypeStr` where being misordered has an effect on functionality.

I've added another commit to fix the order and handled old exports which are troubled.

mrimer (approved):
Thank you!

PR #750, opened 2025-03-08
Add support for building platforms via scripting

Hypexion:
New platforms of both types can now be constructed in rooms, either directly with the Build command, or via build markers. Constructed platforms will merge with existing platform, and this can allow multiple platforms to merge into a single platform.

A minor technical note is that merging platforms can leave behind empty platforms. This isn't entire ideal, but the same thing can happen when building over platforms, so it doesn't appear to have any detrimental effects.


mrimer:
Thanks for the PR, @Hypexion . The code looks good, but I expect this is a change to pre-existing build functionality.

To ensure I'm understanding the change, am I correct in stating that these build options used to not do anything with platforms, and now they will perform here as described? Do we know what risk there is that the new functionality will break existing rooms? What is the anticipated level of impact? How much do we care? etc.

mrimer (approved):

mrimer:
Thanks for clarifying! I'm good with adding the new behavior.

PR #751, opened 2025-03-09
Add togglable movement order hint effects

Hypexion:
Based on the damage and item preview effects for RPG, movement order can now be displayed directly on monsters. It's a key-togglable effect, since it's not information that's always relevant.

Monster remains don't show order, but do get counted (which is needed to match with the order shown when right-clicking). Maybe order should be shown for construct remains, since that can matter. Fegundo ashes do get the effect, since they're counted as alive.

The performance gets a little bad once the monster counter gets high enough, probably due to how many surfaces are getting created. Not sure if that can be streamlined.

### DIFF
--- a/DROD/DrodFontManager.cpp
+++ b/DROD/DrodFontManager.cpp
@@ -530,19 +530,19 @@ bool CDrodFontManager::LoadFonts()
 	this->LoadedFonts[F_ExpandText].OutlineColor = Black;
 	this->LoadedFonts[F_ExpandText].wOutlineWidth = 2;	
 
-	//Load damage preview font.
+	//Load movement order hint font.
 	pFont = GetFont(wstrFont2Filepath, 14);
 	if (!pFont) return false;
-	this->LoadedFonts[F_MovementOrderPreview].pTTFFont = pFont;
-	this->LoadedFonts[F_MovementOrderPreview].wLineSkipHeight = GetFontHeight(F_MovementOrderPreview);
-	this->LoadedFonts[F_MovementOrderPreview].ForeColor = White;
-	this->LoadedFonts[F_MovementOrderPreview].BackColor = Black;
-	this->LoadedFonts[F_MovementOrderPreview].bAntiAlias = true;
-	this->LoadedFonts[F_MovementOrderPreview].bOutline = true;
-	this->LoadedFonts[F_MovementOrderPreview].OutlineColor = Black;
-	this->LoadedFonts[F_MovementOrderPreview].wOutlineWidth = 1;
-	GetWordWidth(F_MovementOrderPreview, wszSpace, wSpaceWidth);
-	this->LoadedFonts[F_MovementOrderPreview].wSpaceWidth = wSpaceWidth;
+	this->LoadedFonts[F_MovementOrderHint].pTTFFont = pFont;
+	this->LoadedFonts[F_MovementOrderHint].wLineSkipHeight = GetFontHeight(F_MovementOrderHint);
+	this->LoadedFonts[F_MovementOrderHint].ForeColor = White;
+	this->LoadedFonts[F_MovementOrderHint].BackColor = Black;
+	this->LoadedFonts[F_MovementOrderHint].bAntiAlias = true;
+	this->LoadedFonts[F_MovementOrderHint].bOutline = true;
+	this->LoadedFonts[F_MovementOrderHint].OutlineColor = Black;
+	this->LoadedFonts[F_MovementOrderHint].wOutlineWidth = 1;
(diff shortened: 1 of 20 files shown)


DROD/DrodFontManager.cpp
```diff
@@ -530,6 +530,20 @@ bool CDrodFontManager::LoadFonts()
 	this->LoadedFonts[F_ExpandText].OutlineColor = Black;
 	this->LoadedFonts[F_ExpandText].wOutlineWidth = 2;	
 
+	//Load damage preview font.
```
mrimer, line 533: Update comment?

DROD/MovementOrderHintEffect.cpp
```diff
@@ -0,0 +1,132 @@
...
+CMovementOrderHintEffect::~CMovementOrderHintEffect()
+{
+	if (this->pTextSurface)
+		SDL_FreeSurface(this->pTextSurface);
+}
+
+//********************************************************************************
+bool CMovementOrderHintEffect::Update(const UINT wDeltaTime, const Uint32 dwTimeElapsed)
+{
+	const UINT wTurnNow = this->pRoomWidget->GetRoom()->GetCurrentGame()->wTurnNo;
```
mrimer, line 72: Suggest add:
ASSERT(this->pRoomWidget->GetRoom());
ASSERT(this->pRoomWidget->GetRoom()->GetCurrentGame());

DROD/MovementOrderHintEffect.h
```diff
@@ -0,0 +1,62 @@
...
+ * Contributor(s):
+ *
+ * ***** END LICENSE BLOCK ***** */
+
+#ifndef MOVEMENTORDERHINTEFFECT_H
+#define MOVEMENTORDERHINTEFFECT_H
+
+#include <FrontEndLib/Effect.h>
+#include <BackEndLib/Types.h>
+#include <BackEndLib/Coord.h>
```
mrimer, line 32: I think the includes for Coord.h and vector can be removed.

DROD/RoomWidget.cpp
```diff
@@ -4734,6 +4747,9 @@ void CRoomWidget::Paint(
 	if (!bPlayerIsAlive) // To ensure everything draws properly during death animation everything must be made dirty
 		this->bAllDirty = true;
 
+	if ((this->bAllDirty || bMoveMade) && bPlayerIsAlive)
```
mrimer, line 4750: I'm wondering whether the performance issue has to do with regenerating them on "bAllDirty". My intuition is that we don't need to rebuild all the effects each time the room is fully refreshed -- only when the turn changes.

The other situation I can think of is when there are no EMOVEORDERHINT effects in the list, but should be. Otherwise, why rebuild them?

Hypexion: The delays are happening between turns being processed, so regenerating on refresh might not be the issue.

A little bit of profiling suggests the heaviest parts are in SDL_tff, but I couldn't get symbols loading for a closer look. Unfortunatey I don't think there's a good way to cut down on the "drawing text" part of this.

mrimer: Thanks for checking. One idea would be to maintain a singleton map caching pre-rendered surfaces, i.e., <turn number, Surface*>. Reuse the surfaces based on the turn number. This should resolve that issue and help avoid a lot of memory address space fragmentation as these surfaces are continuously being deleted and rebuilt.

Hypexion: The caching solution worked out fairly well. I doubt most rooms will get anywhere near a thousand monsters, but it gets a little sluggish around the 300 to 400 mark.

Technically, you can get movement orders over 1216 by using the `Invisible includes move order` on invisible inspectable characters (well, you can now I've bundled in a fix to it with this PR). This is very unlikely in my opinion.

mrimer: Great work! Thank you for implementing the cache. I like how you designed the static clean-up.

Your reply gives me one more thought. Brainstorming, how about we impose an upper bound on the number of possible cached entries? Specifically, what I'm thinking is either:
1. With a ton of monsters in the room, tracking movement order for every one is probably not useful. How about for all entries with a value >= 1000, show the same (cached) symbol, like "+"?
2. Simply don't show an effect for any monster w/ order above a magic number, e.g., >= 1000.

Hypexion: After thinking about it, I decided to have all values >=1000 be disaplayed as `>999`. It's not quite arbitary, since a five character string won't fit on a tile. The exact value is still available through right-click, and if a player needs that information it's perhaps arguable that something has gone wrong somewhere in terms of room design.

mrimer (approved):

PR #765, opened 2025-03-17
RPG: Add commands to alter room lighting

Hypexion:
This PR contains the lighting related changes from #438, but for DROD RPG. This includes the tracking to make it place nicely with undo (but not time travel as RPG doesn't have that).


mrimer (approved):

mrimer:
Thank you!

PR #768, opened 2025-03-19
RPG: Add Attack tile command

Hypexion:
The much desired command to hurt things. It can even be configured to consider or ignore the target's defense value. Percentage damage isn't directly supported, but should be easily achieved with _MyScript overrides. Part of the reason this command is helpful is because #717 means that using `Set monster var` now has side-effects when used on serpents.

(There are also a few things missed from the map icon work in here)

### DIFF
--- a/drodrpg/DRODLib/Character.cpp
+++ b/drodrpg/DRODLib/Character.cpp
@@ -3089,9 +3089,9 @@ void CCharacter::Process(
 			case CCharacterCommand::CC_AttackTile:
 			{
 				getCommandRect(command, px, py, pw, ph);
-				int damage;
+				int damage = 0;
 				CMonster* pMonster = room.GetMonsterAtSquare(px, py);
-				CMonster* pTarget;
+				CMonster* pTarget = NULL;
 				if (pMonster && pMonster->getHP() > 0) {
 					//Attack monster
 					pTarget = pMonster;
@@ -3108,7 +3108,7 @@ void CCharacter::Process(
 					player.Damage(CueEvents, -damage, CID_MonsterKilledPlayer); //-ve for flat damage
 				}
 
-				if (damage == 0) {
+				if (damage == 0 && pTarget) {
 					CueEvents.Add(CID_EntityAffected, new CCombatEffect(pTarget, CET_NODAMAGE), true);
 				}
 
--- a/drodrpg/DRODLib/Character.cpp
+++ b/drodrpg/DRODLib/Character.cpp
@@ -1000,6 +1000,7 @@ void CCharacter::ReflectX(CDbRoom *pRoom)
 			case CCharacterCommand::CC_VarSetAt:
 			case CCharacterCommand::CC_WaitForOpenTile:
 			case CCharacterCommand::CC_SetWallLight:
+			case CCharacterCommand::CC_AttackTile:
 				command->x = (pRoom->wRoomCols-1) - command->x;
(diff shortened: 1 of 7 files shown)


drodrpg/DRODLib/Character.cpp
```diff
@@ -3083,6 +3086,36 @@ void CCharacter::Process(
 				bProcessNextCommand = true;
 			break;
 
+			case CCharacterCommand::CC_AttackTile:
+			{
+				getCommandRect(command, px, py, pw, ph);
+				int damage;
```
mrimer, line 3092: Need to initialize damage=0 and pTarget = NULL for the fallthrough case

drodrpg/DRODLib/Character.cpp
```diff
@@ -3083,6 +3086,36 @@ void CCharacter::Process(
...
+				} else if (pCurrentGame->IsPlayerAt(px, py)) {
+					//Attack player
+					pTarget = &player;
+					int reduction = ph ? 0 : pCurrentGame->getPlayerDEF();
+					damage = pw - reduction;
+					damage = max(0, damage);
+					player.Damage(CueEvents, -damage, CID_MonsterKilledPlayer); //-ve for flat damage
+				}
+
+				if (damage == 0) {
```
mrimer, line 3111: Probably add "&& pTarget != NULL"

drodrpg/DRODLib/CharacterCommand.h
```diff
@@ -364,6 +364,7 @@ class CCharacterCommand
 		CC_SetDarkness,         //Set darkness level in rect (x,y,w,h) to value in flags. Zero value removes darkness
 		CC_SetCeilingLight,     //Set ceiling light value in rect (x,y,w,h) to value in flags. Zero value removes light
 		CC_SetWallLight,        //Set wall light value at (x,y) to intensity (w) with value in flags. Zero intensity or value removes light
+		CC_AttackTile,          //Attack the entity at (x,y) with a power of (w) ATK. Ignores DEF is (h) is set
```
mrimer, line 367: is (h) --> if (h)

drodrpg/Data/Help/1/script.html
```diff
@@ -123,6 +123,9 @@ <h3><u>Character Script Commands</u></h3>
       to orbs before striking them for more intuitive behavior.  If no one is
       next to the orb when struck, it will appear that Beethro is using his
       portable orb.</li>
+  <li><a name="attack-tile"><b>Attack tile</b></a> - Damages the player or monster
+      at the room square you select, with the specified amount of damage. The command
```
mrimer, line 127: maybe reword "amount of damage" --> "amount of ATK power"

drodrpg/Data/Help/1/script.html
```diff
@@ -311,6 +314,11 @@ <h3><u>Character Script Commands</u></h3>
 	  then it will show the room on the minimap as if the player has already
 	  explored it.  If the indicated room has already been explored
 	  by the player, this command does nothing.</li>
+  <li><a name="set-map-icon"><b>Set map icon</b></a> - Adds an icon for the selected
+      room on the player's map. Icons are only shown for rooms that have been revealed
```
mrimer, line 318: Suggest rephrasing for clarity:
Icons can be added to unrevealed rooms but are only displayed on rooms that have been revealed to the player.

PR #775, opened 2025-03-22
RPG: Expand control keybinding options (part one)

Hypexion:
This is part of #327, but for DROD RPG. This PR deals with setting up all the infrastructure needed for setting keybindings that use modifier keys, and sets up the pre-existing editable controls to use the new system. Support isn't being added for non-gameplay commands in this PR, but the commands to use weapons and armor have been broken out to allow them to configured separately from the use accessory command.

Small housekeeping note: Once DROD 5.2 and RPG 2.0 are both merged back into the master branch, we can move `KeypressDialogWidget` back to FrontEndLib.


mrimer (approved):
Thank you!

PR #776, opened 2025-03-22
Fix existance of "drod" directory in repository

Hypexion:
Git and Github treat directory paths in a case-sensitive way. So if you had two directories called `drod` and `DROD`, git will see them as two different directories. Windows treats directory paths in a case-insensitive way, and sees those two directories as being the same.

A developer somehow managed to create a `drod` directory in the repo. This is slightly confusing, so I've taken the steps required to eliminate `drod`.


mrimer (approved):
Thank you!

PR #786, opened 2025-03-26
RPG: Expand control keybinding options (part two)

Hypexion:
The other two halves of #327, but for RPG. Now the editor commands can be changed in RPG. This includes the edit select screen, the room editor screen, the script editing dialog, and also the restore screen. This is a fairly throughout rewiring, but it means we get consistent controls for actions across all contexts. (It also revealed some issues I'll need to fix on the TSS side.)

The specific way the controls are arranged across the two settings tabs can be changed based on feedback.


mrimer (approved):
Thank you!

PR #803, opened 2025-03-30
Fix character logic regressions caused by implementation of behaviors

Hypexion:
A demo test of 5.2 revealed the existance of unwanted changes caused by the implementation of character behaviors.

1. Due to `CCharacter::DoesSquareContainObstacle` being changed to no longer route non-human appearance characters through the `CMonster` version of the function, characters without a dagger weapon became unable to step-kill other monsters. Fixed by checking if the character is human before doing the dagger-step check. (The overall area statement was also cleaned up to be more readable)
2. The same logic change also meant that non-human characters became able to step-kill the player when the player was not a monster target. Fixed by adding a new behavior that controls if a character can move onto a non-target player, which is set by default for human characters.
3. Default behaviors were being set before default movement types whenever those were set. However, the behavior for activating pressure plates is added to non-flying monsters by default. Setting behaviors first meant that flying characters incorrectly activated plated. Fixed by calling `SetDefaultMovementType` before `SetDefaultBehaviors`.
4. Changes to how `CCharacter::IsMonsterTarget` meant that certain character appearances no longer woke evil eyes when the eyes looked at them. A new behavior has been added to make characters wake eyes, which is set by default for characters with a monster target appearance.


mrimer (approved):
Thank you!

PR #807, opened 2025-04-01
RPG: Add setting to enable and disable autosaves

Hypexion:


drodrpg/DRODLib/SettingsKeys.cpp
```diff
@@ -53,6 +53,7 @@ namespace Settings
 	DEF(DamagePreview);
 	DEF(DisableMouse);
 	DEF(DisplayCombos);
+	DEF(EnableAutosave);
```
mrimer, line 56: Is this setting hooked in to game logic anywhere?

Hypexion: It looks like I didn't commit the changes for that. Oops!

PR #824, opened 2025-04-10
RPG: Fix overhead image types and values not being initialised for new rooms

kieranmillar:
Overhead image X and Y were missing the correct type (int) from the database definition. In addition, creating a new room failed to initialize these values correctly, which caused metakit to throw debugasserts whenever loading a room.

Unfortunately, I don't know how to redefine the existing database type, I guess we just need to rebuild the db or something?


mrimer (approved):

mrimer:
Thanks, @kieranmillar !
@Hypexion , at your convenience, would you please confirm whether anything further needs to be done here?

PR #829, opened 2025-04-12
RPG: Add commands to query entire array variables

Hypexion:
#735 for DROD RPG. Also adds the `_ReturnX` predefined var, which previously wasn't required by any commands.


mrimer (approved):

mrimer:
Thank you!

PR #836, opened 2025-04-18
RPG: Make some ability descriptions shorter

Hypexion:
It's nice if ability descriptions aren't so long they take up more than one line on the combat preview dialog. Some suggestions have been made for how they can be more concise.

This PR also makes it so that Goblin and Wyrm weaknesses use the same verbiage as custom weaknesses, for flawless consistency.


mrimer (approved):

mrimer:
Thank you, @Hypexion

PR #838, opened 2025-04-20
RPG: Remove pointless briar distinction in editor and change SFX

kieranmillar:
RPG has both live and dead briar placeable in the editor, but briar's state is entirely dependent upon its connection to a root in RPG, and immediately changes to the correct state on placement. So being able to place both is pointless. Also made it play the briar break sound when placing it.


mrimer (approved):

mrimer:
Thanks!

PR #850, opened 2025-04-30
RPG: Implement incremental XML import and export

Hypexion:
Ports work from #238 and #243 to RPG, which reduce the overall memory requirements for import and export game data by processing data in chunks. (The various fixes for importing uncompressed strings are also included.)


mrimer (approved):
Thank you!

PR #868, opened 2025-05-16
RPG: Support monsters with negative ATK and DEF

Hypexion:
Have negative values for ATK and DEF already works in combat calculations. Since there are some fringe uses for them, I've made it work correctly in other places.

### DIFF
--- a/drodrpg/DRODLib/Monster.cpp
+++ b/drodrpg/DRODLib/Monster.cpp
@@ -1529,12 +1529,12 @@ const
 UINT CMonster::getATK() const
 //Return: monster's ATK
 {
-	UINT val = this->ATK;
+	int val = this->ATK;
 	if (!this->pCurrentGame || !val) //multiply by zero will still be zero, so return now
 		return val;
 	const float fMult = this->pCurrentGame->GetTotalStatModifier(ScriptVars::MonsterATK);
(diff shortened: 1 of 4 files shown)


drodrpg/DRODLib/Monster.cpp
```diff
@@ -1555,15 +1555,16 @@ std::array<float, 3 > CMonster::getHSV() const
...
+	int val = this->DEF;
 	if (!this->pCurrentGame || !val)
 		return val;
 	const float fMult = this->pCurrentGame->GetTotalStatModifier(ScriptVars::MonsterDEF);
-	val = UINTBounds(val * fMult);
-	return !val ? 1 : val; //DEF always >= 1
+	val = intBounds(val * fMult);
+
+	if (val < 0 && IsOnMistTile() && !IsMistImmune())
+		return 0; //Mist tile nullifies +ve DEF
```
mrimer, line 1565: I think the "+ve" is a typo

Hypexion: It's a shorthand but the comment isn't too long. The real typo is that I used the wrong comparator, so mist would stop nullifying positive ATK!

mrimer (approved):

PR #874, opened 2025-05-28
RPG: Fix grappling onto pickup items

kieranmillar:
Fix issue where grappling hook jumps considered some collectables on the target square to be blocking.


mrimer (approved):
I note this logic change could impact 1.2.x holds, like allowing grappling leaps onto maps that weren't allowed before. I'm assuming we're okay with making this change to allow this.

kieranmillar:
This isn't a logic change, it's a fix of a regression. You could grapple onto maps in 1.2.x

PR #875, opened 2025-05-28
RPG: Revert how scripts check player orientation

kieranmillar:
The commands to wait for player to turn clockwise and counterclockwise stopped working after a refactor to check the player's own variables instead of the copy the character was tracking.


mrimer (approved):
Thank you!

PR #876, opened 2025-05-28
RPG: Wait for entity monster should not count the player

kieranmillar:
Fix regression where RPG used to be different from classic DROD in that player roles as monsters should not count when waiting while a monster is at some coordinate.


mrimer (approved):
Thanks!

PR #881, opened 2025-05-31
RPG: Always run full combat simulation for combat dialog

Hypexion:


drodrpg/DRODLib/Combat.h
```diff
@@ -102,7 +102,7 @@ class CCombat
 	bool Advance(CCueEvents& CueEvents, bool bQuickResolution, const UINT maxStrikesToPerform=0);
 
 	bool AttackIsFromBehindMonster(int &dx, int &dy) const;
-	int  GetExpectedDamage();
+	int  GetExpectedDamage(const UINT& maxStrikesToPerform = 0);
```
mrimer, line 105: FYI -- Primitive types don't need to be a const ref; just const is fine. It's the same thing, except references may not be thread safe (not that we're considering this case atm).

PR #882, opened 2025-05-31
RPG: Update hold status for new official hold

Hypexion:
Replaces the `Main` status with `Tendry`, and adds a new `ACR` status. Additionally, where required, checks for the `Main` status have been replaced with `CDbHold::IsOfficialHold`.

### DIFF
--- a/drodrpg/DRODLib/DbHolds.cpp
+++ b/drodrpg/DRODLib/DbHolds.cpp
@@ -1611,6 +1611,26 @@ CDbHold::~CDbHold()
 	Clear();
 }
 
+
+//*****************************************************************************
+//Returns: the official hold status code for this game release
+CDbHold::HoldStatus CDbHold::GetOfficialHoldStatus()
+{
+	return ACR;
+}
+
+//*****************************************************************************
+bool CDbHold::IsOfficialHold(HoldStatus holdstatus)
+{
+	switch (holdstatus) {
+	case Tendry:
+	case ACR:
+		return true;
+
+	default: return false;
+	}
+}
+
 //*****************************************************************************
 UINT CDbHold::AddCharacter(const WCHAR* pwszName)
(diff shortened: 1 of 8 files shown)


drodrpg/DRODLib/DbHolds.cpp
```diff
@@ -3139,8 +3159,15 @@ MESSAGE_ID CDbHold::SetProperty(
...
-			const UINT dwLocalHoldID = GetLocalID(eMatchHoldsWithStatus);
+					? Tendry : NoStatus;	//match player imports for demo and full main hold
+			//Across DLC data, there could be multiple instances of the same author GID.
+			//To join a DLC hold, we need to match against all of these.
+			std::pair<PrimaryKeyMultiMap::const_iterator, PrimaryKeyMultiMap::const_iterator> localIDMap =
+				info.PlayerIDMap.equal_range(this->dwPlayerID);
+			CIDSet localPlayerIDs;
+			for (PrimaryKeyMultiMap::const_iterator it = localIDMap.first; it != localIDMap.second; ++it)
+				localPlayerIDs += it->second;
+			const UINT dwLocalHoldID = GetLocalID(eMatchHoldsWithStatus, localPlayerIDs, this->dwPlayerID);
```
Hypexion, line 3170: Putting a comment here because it's a bit I'm a little unsure about. I did the changes using TSS as a guide, and this is the one place it looks for a specific hold status rather than checking if it's an official hold.

Also, the DLC data checks were added.

mrimer: Ah -- multiple author profiles that are actually the same author (my intuition says this can occur across, say, multiple DLC .dat files that replicate the same player profile as a hold author).

Sure, let's leave this in.

In this case, I think we just need to change "Tendry" to "ACR".

mrimer (approved):

PR #901, opened 2025-06-15
RPG: When loading new style keybinds, save a default if none present

kieranmillar:
When a player is loaded, we run through their key settings and convert them from the old style to the new 5.2 style which allows modifier keys. This change also makes it save down a default setting if the key setting does not exist.

This can happen when a new player is created and hasn't yet opened the settings screen, or the profile is an old one that didn't have all the expanded keybinds that weren't previously configurable.

Due to various refactors to support these new style keybinds, trying to convert these into text for e.g. displaying on scrolls would not work if the key wasn't saved down into the player's settings blob. This change ensures they are always present.

### DIFF
--- a/drodrpg/DROD/DrodScreen.cpp
+++ b/drodrpg/DROD/DrodScreen.cpp
@@ -2451,6 +2451,8 @@ void CDrodScreen::EnablePlayerSettings(
 		return;
 	}
 
+	CDbPlayer::ConvertInputSettings(pPlayer->Settings);
+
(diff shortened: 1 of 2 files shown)


kieranmillar:
While working on something else I still had a breakpoint set, and noticed that this conversion was being run for every single player in the db, which includes all of the authors for imported holds that the average user will never load because they don't run debug builds. This is not only a big waste of time every time we load the player box, looping through ~60 commands per player, but also results in save file bloat, as these players otherwise don't have any keybind settings. I've moved where input conversion is called to only do conversion when trying to play as a player. Hopefully it doesn't break anything.

PR #918, opened 2025-06-21
RPG: Make move counter persist between sessions

Hypexion:
Reported as a bug, although it looks like RPG just didn't support this

Thread: https://forum.caravelgames.com/viewtopic.php?TopicID=46824


mrimer (approved):
Thank you!

PR #922, opened 2025-06-30
RPG: Part 1 of alternative way of tracking map states

kieranmillar:
Part 1 of trying to fix the endless plague of bug reports at map states not persisting correctly across play sessions. I'm splitting this into 2 PRs to get some earlier feedback about the current state of things, and as a backup because I've been working on it for a while. This PR implements everything except saving and loading the saved game.

As outlined at https://forum.caravelgames.com/viewtopic.php?TopicID=46823&page=0#454693 my plan to fix this endless stream of bugs is to create a new saved game whose purpose is to keep track of the best room state a player has seen a room in. This saved game will be saved every time it is updated, so that persistence is guaranteed.

Room map states are updated in 3 occasions:
1) We enter a room not in preview mode (set to Explored)
2) We pick up a map collectible (all relevant rooms set to either NoDetail or Explored accordingly)
3) We use the script command (set to either NoDetail or Explored)

Note that CCurrentGame has a method AddRoomToMap but I have avoided setting state in there so we can make a big efficiency gain when collecting a level map, which would call this once for each room, potentially saving to the DB many many times in a row. Instead I collect the list of rooms to update, update them all together then make one save afterwards.

Room map states are read by:
1) Drawing the minimap
2) Clicking on rooms on the minmap (big and small)
3) Checking if we can scroll to that room when in room preview mode
4) Checking if we can quick travel to the room when its adjacent

### DIFF
--- a/drodrpg/DROD/GameScreen.cpp
+++ b/drodrpg/DROD/GameScreen.cpp
@@ -2447,12 +2447,12 @@ void CGameScreen::OnClick(
 					{
 						ExploredRoom *pExpRoom = this->pCurrentGame->getExploredRoom(roomID);
 						if ((pExpRoom && pExpRoom->HasDetail()) ||
-							this->pCurrentGame->pTotalMapStates->GetStoredMapStateForRoom(roomID) >= MapState::Preview)
+							this->pCurrentGame->GetStoredMapStateForRoom(roomID) >= MapState::Preview)
 						{
 							ShowRoomTemporarily(roomID);
-						}
-						else
+						} else {
 							ToggleBigMap();
+						}
 					}
 				}
 			}
@@ -2494,12 +2494,13 @@ void CGameScreen::OnClick(
 				} else {
 					ExploredRoom *pExpRoom = this->pCurrentGame->getExploredRoom(roomID);
 					if ((pExpRoom && pExpRoom->HasDetail()) ||
-						this->pCurrentGame->pTotalMapStates->GetStoredMapStateForRoom(roomID) >= MapState::Preview)
+						this->pCurrentGame->GetStoredMapStateForRoom(roomID) >= MapState::Preview)
 					{
 						ShowRoomTemporarily(roomID);
 					}
-					else
+					else {
 						ToggleBigMap(); //no room to show -- close map
+					}
 				}
 			}
 			break;
@@ -4370,7 +4371,7 @@ void CGameScreen::SearchForPathToNextRoom(
 		const UINT newRoomID = pRoom->dwRoomID;
 		delete pRoom;
 		if (!this->pCurrentGame->IsRoomExplored(newRoomID) &&
-			this->pCurrentGame->pTotalMapStates->GetStoredMapStateForRoom(newRoomID) != MapState::Invisible)
+			this->pCurrentGame->GetStoredMapStateForRoom(newRoomID) != MapState::Invisible)
 		{
 			this->pRoomWidget->DisplaySubtitle(
 					g_pTheDB->GetMessageText(MID_QuickPathNotAvailable), wPX, wPY, true);
@@ -8654,8 +8655,8 @@ void CGameScreen::DisplayAdjacentTempRoom(const UINT direction)
 
 	const UINT dwNewRoomID = this->pCurrentGame->pLevel->GetRoomIDAtCoords(newRoomX, newRoomY);
 
-	if (!this->pCurrentGame->IsRoomAtCoordsExplored(newRoomX, newRoomY) &&
-		this->pCurrentGame->pTotalMapStates->GetStoredMapStateForRoom(dwNewRoomID) < MapState::Preview) {
+	if (!this->pCurrentGame->CDbSavedGame::IsRoomExplored(dwNewRoomID) &&
+		this->pCurrentGame->GetStoredMapStateForRoom(dwNewRoomID) < MapState::Preview) {
 		g_pTheSound->PlaySoundEffect(SEID_CHECKPOINT);
 		return;
 	}
--- a/drodrpg/DROD/GameScreen.cpp
+++ b/drodrpg/DROD/GameScreen.cpp
@@ -8655,7 +8655,7 @@ void CGameScreen::DisplayAdjacentTempRoom(const UINT direction)
 
 	const UINT dwNewRoomID = this->pCurrentGame->pLevel->GetRoomIDAtCoords(newRoomX, newRoomY);
 
-	if (!this->pCurrentGame->CDbSavedGame::IsRoomExplored(dwNewRoomID) &&
+	if (!this->pCurrentGame->IsRoomExplored(dwNewRoomID) &&
 		this->pCurrentGame->GetStoredMapStateForRoom(dwNewRoomID) < MapState::Preview) {
 		g_pTheSound->PlaySoundEffect(SEID_CHECKPOINT);
 		return;
--- a/drodrpg/DROD/GameScreen.cpp
+++ b/drodrpg/DROD/GameScreen.cpp
@@ -2446,9 +2446,13 @@ void CGameScreen::OnClick(
 					if (!this->pCurrentGame->InCombat() && !this->bNeedToProcessDelayedQuestions)
 					{
 						ExploredRoom *pExpRoom = this->pCurrentGame->getExploredRoom(roomID);
-						if (pExpRoom && pExpRoom->HasDetail())
+						if ((pExpRoom && pExpRoom->HasDetail()) ||
+							this->pCurrentGame->pTotalMapStates->GetStoredMapStateForRoom(roomID) >= MapState::Preview)
+						{
 							ShowRoomTemporarily(roomID);
-						else ToggleBigMap();
+						}
+						else
+							ToggleBigMap();
 					}
 				}
 			}
@@ -2489,9 +2493,13 @@ void CGameScreen::OnClick(
 					ToggleBigMap();
 				} else {
 					ExploredRoom *pExpRoom = this->pCurrentGame->getExploredRoom(roomID);
-					if (pExpRoom && pExpRoom->HasDetail())
+					if ((pExpRoom && pExpRoom->HasDetail()) ||
+						this->pCurrentGame->pTotalMapStates->GetStoredMapStateForRoom(roomID) >= MapState::Preview)
+					{
 						ShowRoomTemporarily(roomID);
-					else ToggleBigMap(); //no room to show -- close map
+					}
+					else
+						ToggleBigMap(); //no room to show -- close map
 				}
 			}
 			break;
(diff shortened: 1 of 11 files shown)


drodrpg/DROD/GameScreen.cpp
```diff
@@ -2446,9 +2446,13 @@ void CGameScreen::OnClick(
 					if (!this->pCurrentGame->InCombat() && !this->bNeedToProcessDelayedQuestions)
 					{
 						ExploredRoom *pExpRoom = this->pCurrentGame->getExploredRoom(roomID);
-						if (pExpRoom && pExpRoom->HasDetail())
+						if ((pExpRoom && pExpRoom->HasDetail()) ||
+							this->pCurrentGame->pTotalMapStates->GetStoredMapStateForRoom(roomID) >= MapState::Preview)
```
mrimer, line 2450: Let's avoid calling into members of the CCurrentGame class where possible and keep the codebase simpler by adding a wrapper method in CCurrentGame that queries pTotalMapStates, e.g., in CurrentGame.h:
MapState CCurrentGame::GetStoredMapStateForRoom(const UINT roomID) { return this->pTotalMapStates->GetStoredMapStateForRoom(roomID); }

Then all the callers can call this one.

drodrpg/DROD/GameScreen.cpp
```diff
@@ -8643,7 +8652,10 @@ void CGameScreen::DisplayAdjacentTempRoom(const UINT direction)
 		default: ASSERT(!"Unsupported temp room pan direction"); return;
 	}
 
-	if (!this->pCurrentGame->IsRoomAtCoordsExplored(newRoomX, newRoomY)) {
+	const UINT dwNewRoomID = this->pCurrentGame->pLevel->GetRoomIDAtCoords(newRoomX, newRoomY);
+
+	if (!this->pCurrentGame->IsRoomAtCoordsExplored(newRoomX, newRoomY) &&
```
mrimer, line 8657: Now that you have dwNewRoomID, you can simplify this check to
CDbSavedGame::IsRoomExplored(dwNewRoomID);
which is the logic that IsRoomAtCoordsExplored includes, and it avoids the now-redundant call to GetRoomIDAtCoords.

drodrpg/DRODLib/TotalMapStates.h
```diff
@@ -0,0 +1,26 @@
...
+#include "DbSavedGames.h"
+#include <BackEndLib/Types.h>
+
+#include <map>
+
+//*******************************************************************************
+class CTotalMapStates {
+public:
+	CTotalMapStates();
+	~CTotalMapStates() { };
```
mrimer, line 13: Don't need a trailing semicolon on a method implementation

drodrpg/DRODLib/TotalMapStates.cpp
```diff
@@ -0,0 +1,142 @@
...
+//*****************************************************************************
+//
+//Public methods.
+//
+
+//*****************************************************************************
+//Constructor
+CTotalMapStates::CTotalMapStates()
+{
+	this->mapStates = map<UINT, MapState>();
```
mrimer, line 44: You don't need this line; mapStates is automatically default constructed to an empty map.

drodrpg/DRODLib/CurrentGame.h
```diff
@@ -399,13 +400,15 @@ class CCurrentGame : public CDbSavedGame
 	CDbLevel *  pLevel;
 	CDbHold *   pHold;
 	CEntranceData *pEntrance;
+	CTotalMapStates *pTotalMapStates;
```
mrimer, line 403: I don't think we need to make this var a pointer, calling new and delete on it here and there. Let's simply make it an object, i.e.,
`CTotalMapStates totalMapStates;`
Now you don't need to worry about memory alloc/dealloc and checking whether the pointer is valid, etc.

I'm assuming that Load(...) will always be called -- not only when allocating a new object, for instance.

drodrpg/DROD/MapWidget.cpp
```diff
@@ -927,9 +927,14 @@ bool CMapWidget::SelectRoomIfValid(
...
+		MapState storedMapState = MapState::Invisible;
+		if (this->pCurrentGame && this->pCurrentGame->pTotalMapStates)
+			storedMapState = this->pCurrentGame->pTotalMapStates->GetStoredMapStateForRoom(dwRoomID);
 		//Has it been explored?
-		if (this->bEditing || this->ExploredRooms.has(dwRoomID) ||
-				this->NoDetailRooms.has(dwRoomID)) //Yes.
+		if (this->bEditing ||
+			this->ExploredRooms.has(dwRoomID) ||
+			this->NoDetailRooms.has(dwRoomID) ||
+			storedMapState != MapState::Invisible) //Yes.
```
mrimer, line 937: PSA: If-clause checks are performed from left to right. As an optimization, we can put the cheaper checks first. Let's put this new check (which might be a single CPU instruction) before the complex data structure scans.

drodrpg/DRODLib/Character.cpp
```diff
@@ -2443,7 +2443,10 @@ void CCharacter::Process(
 					if (roomID)
 					{
 						const bool bMarkExplored = command.w != 0;
-						pGame->AddRoomToMap(roomID, bMarkExplored ? MapState::Explored : MapState::NoDetail);
+						MapState mapState = bMarkExplored ? MapState::Explored : MapState::NoDetail;
```
mrimer, line 2446: I like how you're clarifying the code with these simple refactorings.

Make mapState const for these kinds of var declarations. The logic is easier to reason about when it's explicit that vars are immutable.

drodrpg/DRODLib/CurrentGame.cpp
```diff
@@ -6871,7 +6888,9 @@ void CCurrentGame::ProcessPlayerMoveInteraction(int dx, int dy, CCueEvents& CueE
 		//Level map.
 		//Mark all non-secret rooms in level on map.
 		CIDSet roomsInLevel = CDb::getRoomsInLevel(this->pLevel->dwLevelID);
+		CIDSet roomsWithStatesToUpdate = CIDSet();
```
mrimer, line 6891: You can omit the "= CIDSet()" -- this is a default constructor option, so it's automatically called implicitly.

mrimer (changes requested):
I haven't deeply thought through this approach, but I like your idea about not having to scan through a bunch of saved games to (re)construct a persistent record of room states.

If it's okay, I'd like to do some code cleanup before committing this change (see my comments). This will make it easier for me to review the meat of the logic you're wanting to introduce.

drodrpg/DROD/MapWidget.cpp
```diff
@@ -928,8 +928,11 @@ bool CMapWidget::SelectRoomIfValid(
 	if (dwRoomID) //Yes.
 	{
 		MapState storedMapState = MapState::Invisible;
-		if (this->pCurrentGame && this->pCurrentGame->pTotalMapStates)
-			storedMapState = this->pCurrentGame->pTotalMapStates->GetStoredMapStateForRoom(dwRoomID);
+		{//scoping
+			CCurrentGame* pGame = const_cast<CCurrentGame*>(this->pCurrentGame);
```
mrimer, line 932: Place const_cast inside the inner if block. Shouldn't const cast a null pointer.

drodrpg/DRODLib/TotalMapStates.h
```diff
@@ -9,8 +9,8 @@
 //*******************************************************************************
 class CTotalMapStates {
 public:
-	CTotalMapStates();
-	~CTotalMapStates() { };
+	CTotalMapStates() { }
+	~CTotalMapStates() { }
 	void Load(const UINT dwPlayerID, const UINT dwHoldID);
 	MapState GetStoredMapStateForRoom(const UINT roomID);
```
mrimer, line 15: The trick is to make this method const, and then you can remove the const casts.

drodrpg/DRODLib/CurrentGame.h
```diff
@@ -452,6 +451,12 @@ class CCurrentGame : public CDbSavedGame
 	CCombat  *pCombat;
 	CMonster *pBlockedSwordHit; //when not NULL, indicates this turn's movement is invalid due to a forbidden attack on this monster
 
+	//TotalMapStates
+	CTotalMapStates TotalMapStates;
+	MapState GetStoredMapStateForRoom(const UINT roomID) { return TotalMapStates.GetStoredMapStateForRoom(roomID); }
```
mrimer, line 456: Make method const

drodrpg/DRODLib/TotalMapStates.h
kieranmillar: This made no difference, it still refuses to let me call it from other classes. The only fix seems to be to stop making the CCurrentGame instance in MapWidget a const.

mrimer: > This doesn't work because TotalMapStates itself is not const inside CCurrentGame, and it can't be because we need to be able to update the room states. Even though this isn't relevant to the particular pathway of calling GetStoredMapStateForRoom, the compiler will not be happy unless all the object instances along the way are also const.

I'm probably not following the use cases here, but regarding the instance in MapWidget, if only const methods are called there, then that instance can remain const.

kieranmillar: Sorry, my bad, I did not understand that to make the method const you have to put the const keyword at the end of the definition, not the front.

drodrpg/DRODLib/CurrentGame.h
```diff
@@ -449,6 +451,12 @@ class CCurrentGame : public CDbSavedGame
 	CCombat  *pCombat;
 	CMonster *pBlockedSwordHit; //when not NULL, indicates this turn's movement is invalid due to a forbidden attack on this monster
 
+	//TotalMapStates
+	CTotalMapStates TotalMapStates;
+	const MapState GetStoredMapStateForRoom(const UINT roomID) const { return TotalMapStates.GetStoredMapStateForRoom(roomID); }
```
mrimer, line 456: Remove the preceding const.

drodrpg/DROD/MapWidget.cpp
mrimer: Noticed this comment is still outstanding here. Let's address. After that, do you know how to squash the set of commits into a single commit?

drodrpg/DRODLib/TotalMapStates.h
```diff
@@ -0,0 +1,26 @@
...
+
+#include <map>
+
+//*******************************************************************************
+class CTotalMapStates {
+public:
+	CTotalMapStates() { }
+	~CTotalMapStates() { }
+	void Load(const UINT dwPlayerID, const UINT dwHoldID);
+	const MapState GetStoredMapStateForRoom(const UINT roomID) const;
```
mrimer, line 15: Remove preceding const.

mrimer (changes requested):

PR #923, opened 2025-07-05
RPG: Part 2 of alternative way of tracking map states

kieranmillar:
Following on from #922 , this PR now saves, loads, exports and imports the new way of tracking what rooms we've previously seen. The old method, where we checked every saved game, now only occurs once on importing a player or save, so that RPG 1 player profiles will still be able to retain this functionality when importing their RPG 1 profile.

This new method strongly overlaps with the explored room tally system saved into ST_PlayerTotal, which also maintains one record per room we've explored, but I left that in as I didn't want to rock that boat too much for systems unrelated to this one.

### DIFF
--- a/drodrpg/DROD/GameScreen.cpp
+++ b/drodrpg/DROD/GameScreen.cpp
@@ -2447,7 +2447,7 @@ void CGameScreen::OnClick(
 					{
 						ExploredRoom *pExpRoom = this->pCurrentGame->getExploredRoom(roomID);
 						if ((pExpRoom && pExpRoom->HasDetail()) ||
-							this->pCurrentGame->GetStoredMapStateForRoom(roomID) >= MapState::Preview)
+							CDbSavedGames::IsMoreDetailedMapState(this->pCurrentGame->GetStoredMapStateForRoom(roomID), MapState::NoDetail))
 						{
 							ShowRoomTemporarily(roomID);
 						} else {
@@ -2494,7 +2494,7 @@ void CGameScreen::OnClick(
 				} else {
 					ExploredRoom *pExpRoom = this->pCurrentGame->getExploredRoom(roomID);
 					if ((pExpRoom && pExpRoom->HasDetail()) ||
-						this->pCurrentGame->GetStoredMapStateForRoom(roomID) >= MapState::Preview)
+						CDbSavedGames::IsMoreDetailedMapState(this->pCurrentGame->GetStoredMapStateForRoom(roomID), MapState::NoDetail))
 					{
 						ShowRoomTemporarily(roomID);
 					}
@@ -4371,7 +4371,7 @@ void CGameScreen::SearchForPathToNextRoom(
 		const UINT newRoomID = pRoom->dwRoomID;
 		delete pRoom;
 		if (!this->pCurrentGame->IsRoomExplored(newRoomID) &&
-			this->pCurrentGame->GetStoredMapStateForRoom(newRoomID) != MapState::Invisible)
+			CDbSavedGames::IsMoreDetailedMapState(this->pCurrentGame->GetStoredMapStateForRoom(newRoomID), MapState::Invisible))
 		{
 			this->pRoomWidget->DisplaySubtitle(
 					g_pTheDB->GetMessageText(MID_QuickPathNotAvailable), wPX, wPY, true);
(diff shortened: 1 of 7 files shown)


drodrpg/DRODLib/DbSavedGames.cpp
```diff
@@ -3113,6 +3119,129 @@ void CDbSavedGames::UpdatePlayerTallies(
...
+	CDbSavedGame* pMapStateSave = GetByID(dwID);
+	if (!pMapStateSave)
+		return mapStates;
+
+	CIDSet exploredRooms = pMapStateSave->GetExploredRooms(true, true);
+	exploredRooms.intersect(rooms);
+	CIDSet::const_iterator room;
+	for (room = exploredRooms.begin(); room != exploredRooms.end(); ++room)
+	{
+		UINT roomID = *room;
```
mrimer, line 3138: Make const

drodrpg/DRODLib/DbSavedGames.cpp
```diff
@@ -3113,6 +3119,129 @@ void CDbSavedGames::UpdatePlayerTallies(
...
+		vector<ExploredRoom*>::const_iterator room;
+		for (room = pMapStateSave->ExploredRooms.begin(); room != pMapStateSave->ExploredRooms.end(); ++room)
+		{
+			ExploredRoom* pRoom = *room;
```
mrimer, line 3142: You can double-dereference pointers for greater efficiency on repeat accesses, like this:
`const ExploredRoom& room = **room;`
and then access the values directly like
`if (room.roomID == roomID)`, etc.

drodrpg/DRODLib/DbSavedGames.cpp
```diff
@@ -3113,6 +3119,129 @@ void CDbSavedGames::UpdatePlayerTallies(
...
+	//Merge RoomMapStates list with existing list.
+	bool bUpdate = false;
+	for (RoomMapStates::const_iterator it = RoomMapStates.begin(); it != RoomMapStates.end(); ++it)
+	{
+		const UINT roomID = it->first;
+		ASSERT(roomID);
+		vector<ExploredRoom*>::const_iterator room;
+		for (room = pMapStateSave->ExploredRooms.begin(); room != pMapStateSave->ExploredRooms.end(); ++room)
+		{
+			ExploredRoom* pRoom = *room;
```
mrimer, line 3179: Same here:
`ExploredRoom &room = **room;`

drodrpg/DRODLib/DbSavedGames.cpp
```diff
@@ -3113,6 +3119,129 @@ void CDbSavedGames::UpdatePlayerTallies(
...
+		{
+			for (vector<ExploredRoom*>::const_iterator room = pSavedGame->ExploredRooms.begin(); room != pSavedGame->ExploredRooms.end(); ++room)
+			{
+				ExploredRoom* pRoom = *room;
+				MapState currentMapState = MapState::Invisible;
+				RoomMapStates::const_iterator currentMapStateIter = mapStates.find(pRoom->roomID);
+				if (currentMapStateIter != mapStates.end()) {
+					currentMapState = currentMapStateIter->second;
+				}
+				if (pRoom->mapState > currentMapState)
```
mrimer, line 3233: The ">" operator check feels potentially fragile for long-term maintenance if new values ever get added to MapState.

I'm wondering whether it makes sense to refactor comparisons like this, which implicitly assume higher values are the ones we want to retain, into a method called something like "IsMoreDetailedRoomState" that encapsulates this logic. Reason being is that it may not be true at some point in the future that a greater value always indicates an "improved" room state that we want to track here.

WDYT?

drodrpg/DRODLib/TotalMapStates.cpp
```diff
@@ -27,30 +27,42 @@
...
 //Public methods.
 //
 
 //*****************************************************************************
 //Load the map states from the saved game
-void CTotalMapStates::Load(const UINT dwPlayerID, const UINT dwHoldID)
+void CTotalMapStates::Load(const UINT dwPlayerID, const UINT dwLevelID)
 {
-	//TODO
+	CIDSet roomsInLevel = CDb::getRoomsInLevel(dwLevelID);
```
mrimer, line 44: Make const

drodrpg/DRODLib/DbSavedGames.cpp
kieranmillar: Sounds like a good idea to me. I've made this change.

drodrpg/DROD/GameScreen.cpp
```diff
@@ -2447,7 +2447,7 @@ void CGameScreen::OnClick(
 					{
 						ExploredRoom *pExpRoom = this->pCurrentGame->getExploredRoom(roomID);
 						if ((pExpRoom && pExpRoom->HasDetail()) ||
-							this->pCurrentGame->GetStoredMapStateForRoom(roomID) >= MapState::Preview)
+							CDbSavedGames::IsMoreDetailedMapState(this->pCurrentGame->GetStoredMapStateForRoom(roomID), MapState::NoDetail))
```
mrimer, line 2450: Just checking -- should the second argument still be "Preview"?
(also below)

kieranmillar: These are correct, the map state has to be better than NoDetail in order to remotely view it. The new method does a greater than, the previous code here was greater than or equal to.

PR #962, opened 2025-08-23
RPG: Fix expansion when assigning text using a character local predef…

kieranmillar:
…ined var

If you tried to assign text and it included a reference to a character-local predefined var, such as _MyATK , then this would always return 0 as the value.

https://forum.caravelgames.com/viewtopic.php?TopicID=46963


kieranmillar:
> Would you please rebase this PR to resolve the conflict?

I've resolved the merge conflict, it will be much easier if you can do a squash merge on this PR to strip out the multiple commits.

mrimer:
Thanks. I will squash merge. There is still a message saying there is a conflict to resolve before merging.

mrimer (approved):

PR #981, opened 2025-09-09
Updated Linux build scripts

enzi:
- git workflow for linux
- minor changes to fix compiler errors
- updated ninjamaker file that uses mostly static libs
- build-deps.sh that pulls and builds all dependencies

to build now:
./ninjamaker (options are -caravel or -steam -release -64)
./build-deps.sh
./build

### DIFF
--- a/DROD/Main.cpp
+++ b/DROD/Main.cpp
@@ -146,12 +146,10 @@ char windowTitle[] = GAMETITLE " BETA - DO NOT DISTRIBUTE (buggy releases make u
 char windowTitle[] = GAMETITLE;
 #endif
 
-#ifdef STEAMBUILD
 #ifdef WIN32
 //On Windows, support selecting whether to create and find the user's Data directory
 //in their userspace or in a common location (e.g., "ProgramData")
 bool bWindowsDataFilesInUserSpecificDir = false;
-#endif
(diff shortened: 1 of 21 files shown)


DROD/Main.cpp
```diff
@@ -146,11 +146,13 @@ char windowTitle[] = GAMETITLE " BETA - DO NOT DISTRIBUTE (buggy releases make u
 char windowTitle[] = GAMETITLE;
 #endif
 
+#ifdef STEAMBUILD
```
mrimer, line 149: What impact will this have?

mrimer:
@enzi Thank you for the PR!

DROD/Main.cpp
enzi: hm, good call. i'll go over this again tomorrow. declaring the variable just for win32 and steam looks wrong indeed.
there are occurrences where it's set to false with just a steam directive so it fails to build on any non win32 builds.

enzi: keen eye, wrong place and addressed here: https://github.com/CaravelGames/drod/pull/981/commits/6edf1e249c9d7b4d1e7819b2f3d8a8f09e735eb8

PR #1012, opened 2025-10-23
RPG: Serialize CCharacter::jumpStack

Hypexion:
Currently the return positions for `GoSub` commands aren't being saved, which causes scripts to break if you leave the room while a script is in subroutine.


mrimer:
Thank you, @Hypexion !

PR #1077, opened 2026-02-26
Made steam builds work:

EvidentlyCube:
# Reasonings
1. `-lsteam_api` must be included very late in the list of argument so I had to expand `ninjamaker`to support that
2. Changed linking of SDL2 to be dynamic because otherwise Steam is not able to inject the overlay
3. Had to move `nghttp2` to be a static lib because otherwise Steam injected its own version that was too old and did not work

# Fluff
With these changes I can make Steam builds and non-steam builds and everything seems to be working fine.

### DIFF
--- a/.github/workflows/build-linux.yml
+++ b/.github/workflows/build-linux.yml
@@ -48,6 +48,9 @@ jobs:
           ninja-build \
           libssl-dev \
           libnghttp2-dev \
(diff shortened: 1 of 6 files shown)


Hypexion (approved):
Nothing here looks objectional, and builds appear to be working. Can merge if nothing else needs doing.

PR #1104, opened 2026-04-06
Move KeypressDialogWidget to FrontEndLib

Hypexion:
Since TSS 5.2 and RPG 2 were developed on separate branches, we ended up duplicating the files for `KeypressDialogWidget`. Now that everything's been merged back, we should replace the two game-specific implementations with a single common implementation in FrontEndLib.

The main changes were clearing up the use of `g_pTheDB`, which we don't want to be doing in FrontEndLib. I also added `F_Header` to the list of fonts in `FontManager.h`, as it's used in the widget.


FrontEndLib/KeypressDialogWidget.h
```diff
@@ -37,10 +37,11 @@
 class CKeypressDialogWidget : public CDialogWidget
 {
 public:
-	CKeypressDialogWidget(UINT dwSetTagNo);
+	CKeypressDialogWidget(UINT dwSetTagNo, WSTRING wstrGetKeyText,
```
mrimer: Suggest "const &" on these WSTRINGS (as general practice for non-primitive input params).

mrimer (approved):
Thank you! Feel free to merge at your convenience.

EvidentlyCube:
Checked, no comments from me

PR #1107, opened 2026-04-12
RPG: Include colouring variables in Set monster var

Hypexion:
While the combat stats of multi-tile monsters can be changed with `Set monster var`, they still have to use the standard colouration. By extending `Set monster var`, colour changes can be applied to snakes and rock giants.

The big thing here is moving all the variables that control color up a level to `CMonster`. The way serialisation works means no migration is needed, but I did take the opportunity to do a small refactor.

Example of recoloured large monsters:
image

Thread: https://forum.caravelgames.com/viewtopic.php?TopicID=46933


drodrpg/DRODLib/Character.h
```diff
@@ -128,11 +125,7 @@ class CCharacter : public CPlayerDouble
 	virtual bool   HasNoEnemyDefense() const {return this->bNoEnemyDEF;}
 	virtual bool   HasSerpentWeakness() const {return this->bSerpentWeakness;}
 
-	virtual UINT   getATK() const;   //allow "negative" values to be returned
-	virtual UINT   getColor() const;
-	virtual UINT   getHue() const;
-	virtual UINT   getSaturation() const;
-	virtual std::array<float, 3> getHSV() const;
+	virtual UINT   getATK() const;   //allow "negative" values to be returned;
```
mrimer: Remove comment semicolon

mrimer (approved):
One comment, commenting on a comment

mrimer (approved):

PR #1117, opened 2026-04-22
RPG: Refactor PlayerStats::IsGlobalStatIndex to be readable

Hypexion:
`PlayerStats::IsGlobalStatIndex` is not actually a function a person can look at and understand. The combination of using the index values and inequality checks makes it very difficult to work out which predefined vars will be considered global values. This has a knock-on effect of making it very easy to miss values should should be marked as global. For example, `_ReturnX` is currently not persisted, but it should be.

I've replaced the eldritch logic statement with a big switch statement. Additionally, the function converts the index back to the enumeration value, so that we can use the enums in the switch, thus making the entire function readable. This also fixes `_ReturnX` not being persisted.


EvidentlyCube:
the eldritch is no more!

mrimer (approved):

PR #1122, opened 2026-04-23
Fix command options still visible when filtered out all of them.

EvidentlyCube:
**ISSUE:** In add/edit command dialog when you type a filter out that hides all entries from the actions list whatever option was last selected its details are still visible on the right making it seem like you are editing it (if you for example forget a space between "go to"). When focus on the list is lost it then auto selects "Appear" while still keeping the right-half of the dialog populated with whatever was there earlier.

**DIAGNOSIS:** The select item event handler is not triggered when the filter empties the list; this prevents the dialog from updating itself.

**SOLUTION:** List Box Widget can now be configured to dispatch the event if the list is emptied to select nothing. I did it as an opt-in option to avoid accidentally breaking something in any other existing list.

Reference: https://forum.caravelgames.com/viewtopic.php?TopicID=47352&page=0&message=457778#457778

### DIFF
--- a/FrontEndLib/ListBoxWidget.h
+++ b/FrontEndLib/ListBoxWidget.h
@@ -121,6 +121,7 @@ public:
 	bool           SelectLineWithText(const WCHAR* pText);
 	void           SelectMultipleItems(const bool bVal);
 	void           SetAllowFiltering(const bool bVal) { this->bAllowFiltering = bVal; }
+	void           SetFilterDispatchOnEmpty(const bool bVal) { this->bFilterDispatchOnEmpty = bVal; }
 	void           SetHotkeyItemSelection(const bool bVal=true) {this->bHotkeyItemSelection = bVal;}
 	void           SetItemColor(const UINT dwKey, const SDL_Color& color);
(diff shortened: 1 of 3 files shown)


FrontEndLib/ListBoxWidget.h
```diff
@@ -172,6 +173,8 @@ class CListBoxWidget : public CScrollableWidget
 	UINT           wDraggingLineNo;  //this line # is being moved in the list
 	bool           bRearranged;      //choices were rearranged on mouse drag
 	bool           bAllowFiltering;  //Allow filtering by typing text
+	/// Dispatch change event when filtering and the list becomes empty
+	bool           bFilterDispatchOnEmpty;
```
mrimer, line 177: Init to false?

mrimer (changes requested):

FrontEndLib/ListBoxWidget.h
EvidentlyCube: Thanks, fixed! C++ is hard

mrimer (approved):

PR #1125, opened 2026-04-24
RPG: Add additional array variable utility script commands

Hypexion:
A few script commands to make working with arrays a little smoother. It also adds `_ReturnY` so that one of them works.

- `Push to array var` allows values to be appending onto the end of an array without having to track where it currently ends. Like `Set array var`, multiple values can be added with a single command.
- `Pop from array var` gets the value on the end of an array and moves it to `_ReturnX`. Then it erases the entry.
- `Get array range` finds the highest and lowest indices in the array, and puts them in `_ReturnX` and `_ReturnY` respectively.

Help content has also been updated.

### DIFF
--- a/drodrpg/Data/Help/1/scriptarray.html
+++ b/drodrpg/Data/Help/1/scriptarray.html
@@ -32,9 +32,9 @@
         <li><a name="count-entries"><b>Count array entries</b></a> - Counts the number of values in the array that satisfies the given comparison to a given value.
         The result is stored in <b>_ReturnX</b>.</li>
     </ul></p>
-    <p>Additionally, the <a name="array-range"><b>Get array var range</b></a> command can be used to find an array's highest and lowest index. The highest index among set
-    values is stored in <b>_ReturnX</b>, and the lowest index among set values is stored in <b>_ReturnY</b>. If this command is used on an empty array, <b>_ReturnX</b> will be set to -1,
-    and <b>_ReturnY</b> to 1.</p>
+    <p>Additionally, the <a name="array-range"><b>Get array var range</b></a> command can be used to find an array's highest and lowest index. The lowest index among set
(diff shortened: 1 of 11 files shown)


drodrpg/Data/Help/1/scriptarray.html
```diff
@@ -31,8 +31,19 @@ <h3><u>Using Arrays in Scripts</u></h3>
         As with other variable commands, an expression can be used in place of a constant value.</li>
         <li><a name="count-entries"><b>Count array entries</b></a> - Counts the number of values in the array that satisfies the given comparison to a given value.
         The result is stored in <b>_ReturnX</b>.</li>
+    </ul></p>
+    <p>Additionally, the <a name="array-range"><b>Get array var range</b></a> command can be used to find an array's highest and lowest index. The highest index among set
+    values is stored in <b>_ReturnX</b>, and the lowest index among set values is stored in <b>_ReturnY</b>. If this command is used on an empty array, <b>_ReturnX</b> will be set to -1,
+    and <b>_ReturnY</b> to 1.</p>
```
mrimer: Just curious -- why not set return X/Y to 0,0 on an empty array?

Edit: Ah, I see the point is that the range is impossible. It's confusing that the range is returned with the highest (end) value first and lowest (start) value second. I would expect (x,y) to return (start, end), like all math ranges I can think of get returned in programming languages, not (end, start).

mrimer (changes requested):
These are nice new array commands! Thank you.

I left a comment describing how I think the return (X,Y) values should be swapped. WDYT?

drodrpg/Data/Help/1/scriptarray.html
Hypexion: In hindsight, it does make more sense for X to be lowest and Y to be highest. I've switch it around. (And added a clarification to the help that the (1, -1) case is imposible for non-empty arrays)

mrimer (approved):

PR #1128, opened 2026-04-26
RPG: Fully support multiple scorepoints in single turn

Hypexion:
DROD RPG technically supports having multiple scorepoints trigger on a single turn, in the sense that each one will pop up a score dialog and register a local highscore. However, there are two problems:

- All scores are currently calculated at the end of a turn, so all scorepoints from the same turn will report the same score.
- Score verification can only handle one scorepoint at a time. This means a CNet-published hold that triggers multiple scorepoints on a single turn will have scorepoints that can't be verified.

To fix this, I've made some changes to the scorepoint event:

1. I've created a new utility class, `ScoreCheckpointData`, which allows a scorepoint name and a set of player stats to be attached to a `CID_ScoreCheckpoint` event. This replaces the previous situation where only the name was attached.
2. When a `CC_ScoreCheckpoint` command is processed, it can now attach the "in the moment" player stats to a scorepoint.
3. Every place that proccesses a `CID_ScoreCheckpoint` event now uses the attached player stats where stats are required.
4. `CDb::ValidateSavedGame` can now output multiple scorepoint data objects, allowing all scorepoints from a turn to be verified.

This should probably be looked over by @schikore since he'll have to implement server-side changes to consume the updated validation output.

### DIFF
--- a/drodrpg/DRODLib/CueEvents.h
+++ b/drodrpg/DRODLib/CueEvents.h
@@ -198,7 +198,7 @@ enum CUEEVENT_ID
 
 	//A score checkpoint is triggered.
 	//
-	//Private data: CDbMessageText *pScoreIDText (one)
+	//Private data: ScoreCheckpointData *pScoreData (one)
 	CID_ScoreCheckpoint,
 
 	//If any calls to CDbRoom::Plot() were made in the current room, this event will
--- a/drodrpg/DRODLib/CurrentGame.h
+++ b/drodrpg/DRODLib/CurrentGame.h
@@ -192,7 +192,7 @@ struct TarstuffStab {
 //*******************************************************************************
 class ScoreCheckpointData : public CAttachableObject {
 public:
-	ScoreCheckpointData(PlayerStats ps, WSTRING name)
+	ScoreCheckpointData(const PlayerStats& ps, const WSTRING& name)
 		: stats(ps), scorepointName(name) {}
 	~ScoreCheckpointData() {}
 	PlayerStats stats;
--- a/drodrpg/DRODLib/CurrentGame.h
+++ b/drodrpg/DRODLib/CurrentGame.h
@@ -189,6 +189,16 @@ struct TarstuffStab {
 	CMonster* pTarstuffMonster;
 };
 
+//*******************************************************************************
+class ScoreCheckpointData : public CAttachableObject {
+public:
+	ScoreCheckpointData(PlayerStats ps, WSTRING name)
+		: stats(ps), scorepointName(name) {}
+	~ScoreCheckpointData() {}
+	PlayerStats stats;
+	WSTRING scorepointName;
+};
+
 typedef pair<ScriptVars::MapIcon, ScriptVars::MapIconState> MapIconPair;
 
 //*******************************************************************************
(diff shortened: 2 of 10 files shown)


drodrpg/DRODLib/CueEvents.h
```diff
@@ -198,7 +198,7 @@ enum CUEEVENT_ID
 
 	//A score checkpoint is triggered.
 	//
-	//Private data: CDbMessageText *pScoreIDText (one)
+	//Private data: ScoreCheckpointData *pScoreData (one)
```
mrimer, line 201: Is "(one)" still correct?

drodrpg/DRODLib/CurrentGame.h
```diff
@@ -189,6 +189,16 @@ struct TarstuffStab {
 	CMonster* pTarstuffMonster;
 };
 
+//*******************************************************************************
+class ScoreCheckpointData : public CAttachableObject {
+public:
+	ScoreCheckpointData(PlayerStats ps, WSTRING name)
```
mrimer: Make args const &

drodrpg/DRODLib/Db.cpp
```diff
@@ -455,16 +453,16 @@ const
...
-		ASSERT((const WCHAR*)(*pScoreIDText));
-		scoreCheckpointName = (const WCHAR*)(*pScoreIDText);
-		ps = pGame->pPlayer->st;
-
-		//Alter stats to match how they are scored.
-		ps.ATK = pGame->getPlayerATK();
-		ps.DEF = pGame->getPlayerDEF();
+		//Clear any data from previous turns
+		scoresData.clear();
+		for (const CAttachableObject* pObj = CueEvents.GetFirstPrivateData(CID_ScoreCheckpoint);
```
mrimer, line 458: Review caveat: I didn't check all the changed spots to confirm that this is the only spot where the CueEvent objects need to be iterated over.

drodrpg/DRODLib/Db.cpp
```diff
@@ -455,16 +453,16 @@ const
...
-			CueEvents.GetFirstPrivateData(CID_ScoreCheckpoint));
-		ASSERT((const WCHAR*)(*pScoreIDText));
-		scoreCheckpointName = (const WCHAR*)(*pScoreIDText);
-		ps = pGame->pPlayer->st;
-
-		//Alter stats to match how they are scored.
-		ps.ATK = pGame->getPlayerATK();
-		ps.DEF = pGame->getPlayerDEF();
+		//Clear any data from previous turns
+		scoresData.clear();
```
mrimer, line 457: Assuming scoresData is always empty when calling ValidateMoveSequence. If so, not sure the clear() is needed here. Maybe always clear, even if CID_ScoreCheckpoint is not set? Not sure offhand whether this would have unintended side effects.

mrimer (commented):
Thank you! It looks okay. Made some suggestions/questions/comments.

drodrpg/DRODLib/CueEvents.h
Hypexion: It should be. The way things are set up is that each `CC_ScoreCheckpoint` will generated a new `CID_ScoreCheckpoint` event. Each event therefore holds a single `ScoreCheckpointData` object.

drodrpg/DRODLib/Db.cpp
Hypexion: Currently, `ValidateMoveSequence` will pick up each scorepoint that gets achieved along a move sequence, with each one encountered replaced the last as the output. Clearing the output vector replicates this behaviour, while allowing multiple scores to be reported in a single turn.

mrimer (approved):

PR #1129, opened 2026-04-28
Fix issues with text selection and scrolling.

EvidentlyCube:
(In this commit message when I mention "selection pixels" I am referring to the four `nSelect...` properties in `TextBox2DWidget` which are responsible for determining the pixel positions of the first and last selection character that are then used to draw the selection area)

ISSUE: A bunch of issues with text selection and scrolling:
 - When selection contains an empty line, scrolling with mouse wheel or scroll bar buttons causes assertion beep
 - Generally scrolling with mouse wheel breaks selection
 - Scrolling with scroll bar causes issues with selection drawing outside the widget if the selection is "below" the widget

DIAGNOSIS: Basically there were two mechanism for calculating the selection area on the screen:
1. First a selection can be made by arrows/mouse by dragging. This stores the selection pixels in the four `nSelect...` variables.
2. Then if you use the scrollbar it just moves the selection Y by pixels up and down.
3. But if then used mouse wheel or the buttons this caused a call to `CalculateHighlightedRegion()` which regenerated the selection pixels and was broken. It also tried to crop the selected area to the visible area on the screen which meant that if this cropped the selection area and you switched to using the scrollbar the selection would be wrong. It would be wrong if it worked at all.

Then there was only partial cropping during the render phase which had a few issues of its own.

SOLUTION: A big rewrite of major parts on how the selection pixels are calculated. The gist is that we now keep the selection intact as calculated during selecting and then crop it to the visible area during render.

It was the broken selection-pixel-regeneration method. It was called in a few places:
 - `ScrollDownOneLine` and `ScrollUpOneLine` - in both cases it was completely unnecessary because they both call `MoveView...` functions which already move the selection. Removing it fixed the beep and all issues with scrolling with mouse wheel. It still caused a bunch of other issues when selection was fully or partially outside the drawing area.
 - `SelectAllText` - it didn't work correctly and was replaced with a mechanism to calculate the position of the first and last character in the text box.

I tried to salvage the existing highlight drawing code but beyond the general idea it was rewritten from scratch as nothing worked quite well. It now calculates the area of each strip in relation to current view and then crops it to the visible area. This way we don't need special hacks for various scenarios (is the first strip still visible? last strip? what if they are only visible partially?) and if a strip is partially visible we render it partially.

Also disabled optimizations for linux debug builds because that caused too much inlining which made me waste an hour or two chasing the bug in an incorrect place.

Reference: https://forum.caravelgames.com/viewtopic.php?TopicID=47367&page=0

### DIFF
--- a/FrontEndLib/BitmapManager.h
+++ b/FrontEndLib/BitmapManager.h
@@ -110,7 +110,7 @@ public:
 		{BAndWRect(x,y,CX_TILE,CY_TILE,pDestSurface);}
 	void        BlitAlphaSurfaceWithTransparency(SDL_Rect src, SDL_Surface *pSrcSurface,
 			SDL_Rect dest, SDL_Surface *pDestSurface, const Uint8 opacity);
-	void        BlitRGBAtoRGBA(SDL_Surface *pSrcSurface, SDL_Rect *src, 
+	void        BlitRGBAtoRGBA(SDL_Surface *pSrcSurface, SDL_Rect *src,
 			SDL_Surface *pDestSurface, SDL_Rect *dest);
 	void        BlitSurface(SDL_Surface *pSrcSurface, SDL_Rect* src,
 			SDL_Surface *pDestSurface, SDL_Rect* dest,	const Uint8 nOpacity=255);
@@ -140,6 +140,7 @@ public:
 	void        BlitWrappingSurface(SDL_Surface *pSrcSurface, SDL_Rect src,
 			SDL_Surface *pDestSurface, SDL_Rect dest, const Uint8 nOpacity=255);
 
+	static void CropRect(SDL_Rect& cropped, const SDL_Rect& bounds);
 	static void ClipSrcAndDestToRect(SDL_Rect& src, SDL_Rect& dest,
 			const UINT W, const UINT H);
 	static SDL_Surface* ConvertSurface(SDL_Surface *pSurface);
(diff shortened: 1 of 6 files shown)


EvidentlyCube:
@Hypexion since it's all FrontEndLib changes no need to port anything to RPG, should work out of the box

EvidentlyCube:
[Screencast_20260428_134435.webm](https://github.com/user-attachments/assets/fa8359ba-c54c-4045-8e4b-79194afa9a6a)

EvidentlyCube:
Found some more issues, mentioned them in the forum post, so I am back to the fixing board with this

EvidentlyCube:
A bunch of additional fixes with multiline text boxes:

1. Fixed issues with selection not working correctly when using Shift+Page Down/Up and when using Shift+Up/Down when the cursor was right by the top/bottom edge.
2. Page Down no longer scrolls past what is possible to scroll with keyboard/scrollbar
3. Also Page Down now correctly gets to the end of text when used on a widget without scrolling
4. Ctrl+Home/End behave as expected
5. Undoing no longer scrolls the view up so that the edited line is at the bottom of the window
6. Backspace & Delete button actions can now be undone

Ready for review/merge now

FrontEndLib/BitmapManager.h
```diff
@@ -140,6 +140,7 @@ class CBitmapManager
 	void        BlitWrappingSurface(SDL_Surface *pSrcSurface, SDL_Rect src,
 			SDL_Surface *pDestSurface, SDL_Rect dest, const Uint8 nOpacity=255);
 
+	static void CropRect(SDL_Rect& cropped, const SDL_Rect& bounds);
```
mrimer: Though the signature technically matches, the first param name differs from the implementation.

EvidentlyCube: Thanks, fixed!

mrimer (approved):

PR #1131, opened 2026-04-28
Add way to quietly run Build command.

EvidentlyCube:
This is achieved in a backwards-compatible manner to 5.2 by using the unused `label` property of the Build command. It is done in a generic way which stores a string modifier `quiet` in the field. If we ever want/need to expand it to support more things this can be easily achieved by using a delimiter for multiple options.

When Build commands executes with this label set, `BuildUtil` has a static field set to perform quiet building. I decided to do it like this primarily to avoid polluting the already large list of arguments that get passed down. One could argue this allows for a bit more versatility in different places should this ever be required.

When building with quiet build then the event `CID_ObjectBuilt` gets attached `T_SILENT_BUILD` instead of the `baseTile`. The attached value is not used anywhere so it's a safe change. In the future if we ever need those values we'll need to add a new struct/class to attach, possibly with the built position as well.

On the frontend side handling `CID_ObjectBuilt` was slightly changed to check if there is any non-T_SILENT_BUILD private data attached. If yes - play the sound and bail out of the loop. For old scripts this means there will be minimal slowness because the first attached data will always trigger playing the sound.

Reference: https://forum.caravelgames.com/viewtopic.php?TopicID=47369&page=0#457834

### DIFF
--- a/DRODLib/DbBase.cpp
+++ b/DRODLib/DbBase.cpp
@@ -757,6 +757,7 @@ const WCHAR* CDbBase::GetMessageText(
 //		case MID_No: strText = "&No"; break;
 		case MID_Explosives: strText = "Explosives"; break;
 		case MID_PushableObjects: strText = "Pushable Objects"; break;
+		case MID_BuildModifierQuiet: strText = "Quietly"; break;
 		default: break;
 	}
 	if (!strText.empty() && (Language::GetLanguage() == Language::English))
--- a/drodrpg/Texts/Speech.uni
+++ b/drodrpg/Texts/Speech.uni
@@ -293,7 +293,7 @@ Enemy sneak attack
 [MID_HitObstacle]
 [eng]
 Player bumps obstacle
-	
+
 [MID_ItemUsed]
 [eng]
 Item used
@@ -1831,3 +1831,7 @@ Count array entries
 [MID_RemoveTransparentLayer]
 [eng]
 Remove transparent layer object
+
+[MID_BuildModifierQuiet]
+[eng]
+Build modifier Quiet
--- a/drodrpg/Texts/Speech.uni
+++ b/drodrpg/Texts/Speech.uni
@@ -1834,4 +1834,4 @@ Remove transparent layer object
 
 [MID_BuildModifierQuiet]
 [eng]
-Build modifier Quiet
+Quietly
--- a/DROD/CharacterDialogWidget.cpp
+++ b/DROD/CharacterDialogWidget.cpp
@@ -195,6 +195,9 @@ const UINT TAG_ENTITY_LISTBOX = 863;
 const UINT TAG_PLAYER_STATE_LISTBOX = 862;
 const UINT TAG_PLAYER_STATE_LISTBOX2 = 861;
 const UINT TAG_RESTRICTED_LABEL = 860;
+/// Playing sound effect for Build command
+const UINT TAG_BUILD_SOUNDLABEL = 859;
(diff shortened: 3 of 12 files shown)


EvidentlyCube:
# How it looks in the editor:
image

Screenshot_20260428_181724

# How Ctrl+B looks

```
  Build Bomb,27,16,27,16,quiet
  Build Bomb,19,22,19,22
```

# 5.2
No one should really be playing 5.2 once 5.2.1 is out due to two pretty severe crashes and a bunch of other smaller and bigger usability issues BUT! This does not expand on anything and uses the label fields. Meaning loading holds using this into 5.2 will just quietly ignore this label.

EvidentlyCube:
@Hypexion & @mrimer Thoughts on this? I admit it's a selfish change because in my upcoming hold I do a lot of cool stuff with Build command and being able to quiet it would make the experience more pleasant.

And frankly it is something that annoyed me immensely for years :D.

DROD/EditSelectScreen.cpp
```diff
@@ -2565,11 +2565,11 @@ void CEditSelectScreen::PopulateHoldListBox()
 #endif
 		{
 			this->pHoldListBoxWidget->AddItem(pHold->dwHoldID, pHold->NameText,
-#ifdef ENABLE_CHEATS
+// #ifdef ENABLE_CHEATS
 				false
-#else
-				!g_pTheDB->Holds.PlayerCanEditHold(pHold->dwHoldID)
-#endif
```
Hypexion, line 2572: I don't think you meant to commit this change.

drodrpg/Texts/Speech.uni
```diff
@@ -1831,3 +1831,7 @@ Count array entries
 [MID_RemoveTransparentLayer]
 [eng]
 Remove transparent layer object
+
+[MID_BuildModifierQuiet]
+[eng]
+Build modifier Quiet
```
Hypexion: I'm fairly sure not having a newline at the end of the uni file will make DRODUtil upset when it's time to process the strings.

DROD/EditSelectScreen.cpp
EvidentlyCube: Nope. Probably misclicked in the stupid, laggy and awfully slow Git interface in VS Code.

drodrpg/Texts/Speech.uni
EvidentlyCube: Fixed

EvidentlyCube:
I did a soft reset and force pushed the changes to avoid polluting history for that Edit screren file. Hopefully github won't explode.

EvidentlyCube:
# And how the test hold looks without the changes from this PR

Screenshot_20260428_211902

The bombs build as before as well, with sound.

DRODLib/DbBase.cpp
```diff
@@ -757,6 +757,7 @@ const WCHAR* CDbBase::GetMessageText(
 //		case MID_No: strText = "&No"; break;
 		case MID_Explosives: strText = "Explosives"; break;
 		case MID_PushableObjects: strText = "Pushable Objects"; break;
+		case MID_BuildModifierQuiet: strText = "Quietly"; break;
```
mrimer, line 760: Why does this not match the value in speech.uni?

EvidentlyCube: I have honestly no idea why the text in speech.uni was so weird. Fixed it there, this is the correct one.

mrimer (approved):

PR #1133, opened 2026-04-29
Allow using labels as comments.

EvidentlyCube:
Complex scripts can benefit greatly from having comments. It's possible to use labels for this but it has a few issues:
1. Lack of indent makes following the flow of the script harder
2. It's more difficult, at a glance, to differentiate usable labels from comments
3. They pollute the Goto lists when using GoTo/Sub/Answer options

This change makes it so that if a label begins with "//" it is treated like a comment:
- Indents the same way as the rest of the code
- Has gray color
- Does not appear in the goto lists

Reference: https://forum.caravelgames.com/viewtopic.php?TopicID=47371

### DIFF
--- a/DROD/CharacterDialogWidget.cpp
+++ b/DROD/CharacterDialogWidget.cpp
@@ -2598,7 +2598,7 @@ void CCharacterDialogWidget::OnKeyDown(
 				command != this->commandBuffer.end(); ++command)
 			{
 				CCharacterCommand *pCommand = *command;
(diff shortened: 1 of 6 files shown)


EvidentlyCube:
# Here is how it looks in action:

### Comment labels formatting

image

### Comment labels not appearing in the list

image

### Even after renaming to the loop jump is maintained

image

mrimer (approved):

PR #1134, opened 2026-04-29
RPG: Improvements to building and waiting for items

Hypexion:
A few changes to make building items and querying them a little better:

1. The `Build` command and the `Wait for item` command were using the same input widgets, meaning they had to share options. I've split this into two, one for building and one for waiting.
2. Scrolls and platforms are now options available for `Wait for item`. Obstacles have been removed from `Build` because they can't be built. All the special removal virtual tiles have been removed from `Wait for item` because it doesn't make sense to wait for them.
3. Virtual tiles have been added for each token, and they've all been added to both lists.
4. Two new groups have been added to `Wait for item group` - `Any token` and `Any T_layer object`.

Thread: https://forum.caravelgames.com/viewtopic.php?TopicID=47221

### DIFF
--- a/drodrpg/DRODLib/CharacterCommand.h
+++ b/drodrpg/DRODLib/CharacterCommand.h
@@ -235,6 +235,8 @@ namespace ScriptFlag
 		IG_Equipment = 35, //Any equipment slot
 		IG_Keys = 36, //Any key
 		IG_Collectable = 37, //Any item the player can pick up
+		IG_AnyToken = 38, //Any token
+		IG_AnyTTile, //Any T-layer object
 		ItemGroupCount //Total number of defined groups
 	};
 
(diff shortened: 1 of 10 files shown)


drodrpg/DRODLib/CharacterCommand.h
```diff
@@ -235,6 +235,8 @@ namespace ScriptFlag
 		IG_Equipment = 35, //Any equipment slot
 		IG_Keys = 36, //Any key
 		IG_Collectable = 37, //Any item the player can pick up
+		IG_AnyToken = 38, //Any token
+		IG_AnyTTile, //Any T-layer object
```
mrimer: Suggest explicitly assigning to 39 as the value must never change

mrimer (approved):

PR #1141, opened 2026-05-01
Fix clones pressing plates when role change from flying to flying.

EvidentlyCube:
Quick story:
 - TCB added clones & player roles.
 - WHen player role changed from flying to non-flying then clones would press plates.
 - For 5.0.2 I [Maurycy] was fixing a bug with seep and plates and broke that behavior - from that point role change from flying to non-flying clones don't press plates; instead they incorrectly press when changing from flying to flying
- This fixes this behavior but does not restore the pre-5.0.2 clone pressing plates for three reasons: A) This makes it consistent with player behavior
B) It worked like that for more time than the old behavior which makes it more canonical C) There is a bug request about changing the behavior for player role which, if implemented, will need to be made as an expansion to Set Player Appearance command and then we could make them work the same.

COPIED OVER IMPORTANT COMMENT FROM CODE IN THIS COMMIT: Intentionally left [some code] as a future reference for 5.3 release if/when we
	decide to tackle the issue that player changing role from flying to non-flying
	does not cause pressure plate pressing. Nor any other action or interaction.
	This extends to an issue with clones. Prior to TSS 5.0.2 clones would press plates
	if player role changed from flying to non-flying but as part of fixes to a seep
	bug I accidentally broke it so that they only press the plate if they cannot
	press it.

Reference threads:
 - Seep bug: https://forum.caravelgames.com/viewtopic.php?TopicID=39739
 - Player role issue: https://forum.caravelgames.com/viewtopic.php?TopicID=47375
 - Clone issue: https://forum.caravelgames.com/viewtopic.php?TopicID=47180

### DIFF
--- a/DRODLib/CurrentGame.cpp
+++ b/DRODLib/CurrentGame.cpp
@@ -4800,6 +4800,24 @@ void CCurrentGame::SynchClonesWithPlayer(CCueEvents& CueEvents)
 
 	SetCloneWeaponsSheathed();
 
+	/*
+	MAURYCY (SKELL):
+	Intentionally left here as a future reference for 5.3 release if/when we
+	decide to tackle the issue that player changing role from flying to non-flying
+	does not cause pressure plate pressing. Nor any other action or interaction.
+	This extends to an issue with clones. Prior to TSS 5.0.2 clones would press plates
+	if player role changed from flying to non-flying but as part of fixes to a seep
+	bug I accidentally broke it so that they only press the plate if they cannot
+	press it.
+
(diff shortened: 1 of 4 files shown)


EvidentlyCube:
@mrimer and @Hypexion I posted this as a draft for now just so it doesn't get merged until the community has a chance to tell me  (in this thread)[https://forum.caravelgames.com/viewtopic.php?TopicID=47375] that there are unforeseen consequences to this change. I'll change it to non-draft Saturday evening my time if no objections happen.

PR #1146, opened 2026-05-05
Various small linux build improvements.

EvidentlyCube:
- Made it so that the various build commands always make sure docker is up.
- Made tests run outside of docker.
- Added separate commands for building and running tests (in the future it might be a good idea to make a new, separate ninja file only for tests)
- Added docs on how to work with DROD in CLion on Linux

### DIFF
--- a/Scripts/Linux/build.sh
+++ b/Scripts/Linux/build.sh
@@ -26,8 +26,8 @@ Actions:
     build-deps        Build project dependencies inside the docker container
     build-drod        Build DROD (Master/Linux) inside the docker container
     build-drod-rpg    Build DROD RPG (drodrpg/Master/Linux) inside the docker container
-    run-tests         Build DROD tests
-    build-tests       Run DROD tests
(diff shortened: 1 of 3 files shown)


Scripts/Linux/build.sh
```diff
@@ -26,7 +26,9 @@ Actions:
     build-deps        Build project dependencies inside the docker container
     build-drod        Build DROD (Master/Linux) inside the docker container
     build-drod-rpg    Build DROD RPG (drodrpg/Master/Linux) inside the docker container
-    run-tests         Build & run DROD tests
+    run-tests         Build DROD tests
```
mrimer: Swap the descriptions?

EvidentlyCube: Funny, I also had those swapped in the actual bodies at some point and fixed it there but somehow missed the descriptions. Fixed!

mrimer (approved):

PR #1153, opened 2026-05-11
Add more detailed assert option

Hypexion:
Adds a new kind of assert, `ASSERT_DETAIL`. This allows an extra piece of information (currently a raw string or unsigned int) to be spliced into an assert message for more detailed reporting.

As a live example, I've changed several asserts in `DbPackedVars` to use this new assert type.


mrimer:
What is the purpose of this change?

Hypexion:
> What is the purpose of this change?

Had a think about the goal, which is to allow for the possibility of more detailed asserts. This led me to do a completely different solution.

mrimer (approved):

PR #1156, opened 2026-06-15
RPG: Add Skeleton guard feature

Hypexion:
Similar to room lock in TSS, the skeleton key guard is a toggleable meta-state that blocks moves that would spend a skeleton key to open or close a door.

When active, a cross symbol is drawn over the skeleton key icon on the sidebar:
image

I will make a PR in the RPG data repo to add the new tile image required for this feature.

### DIFF
--- a/drodrpg/DROD/GameScreen.cpp
+++ b/drodrpg/DROD/GameScreen.cpp
@@ -680,6 +680,12 @@ void CGameScreen::RedrawStats(
 		pLabel->RequestPaint();
 	}
 
+	if (this->pCurrentGame->bSkeletonKeyGuard) {
+		CLabelWidget* pLabel = DYN_CAST(CLabelWidget*, CWidget*, GetWidget(TAG_SKEY));
+		g_pTheDBM->BlitTileImagePart(TI_SMALL_CROSSES, pLabel->GetX() - 2, pLabel->GetY() - 18,
+			0, 0, 22, 22, pDestSurface);
+	}
+
 	//Draw tile image.
 	for (i=0; i<numMenuPics; ++i)
 	{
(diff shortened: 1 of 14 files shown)


drodrpg/DRODLib/CurrentGame.cpp
```diff
@@ -5416,8 +5427,13 @@ bool CCurrentGame::KnockOnDoor(CCueEvents& CueEvents, const UINT wX, const UINT
 				--ps.greenKeys;
 				CueEvents.Add(CID_ItemUsed, new CMoveCoord(wX, wY, GreenKey), true);
 			} else if (ps.skeletonKeys) {
-				--ps.skeletonKeys;
-				CueEvents.Add(CID_ItemUsed, new CMoveCoord(wX, wY, SkeletonKey), true);
+				if (this->bSkeletonKeyGuard) {
```
mrimer: It probably makes sense to refactor the duplicate code into a unified helper function.

drodrpg/DROD/GameScreen.cpp
```diff
@@ -680,6 +680,12 @@ void CGameScreen::RedrawStats(
 		pLabel->RequestPaint();
 	}
 
+	if (this->pCurrentGame->bSkeletonKeyGuard) {
+		CLabelWidget* pLabel = DYN_CAST(CLabelWidget*, CWidget*, GetWidget(TAG_SKEY));
+		g_pTheDBM->BlitTileImagePart(TI_SMALL_CROSSES, pLabel->GetX() - 2, pLabel->GetY() - 18,
```
mrimer: The plural name indicates an interesting use of including multiple images in one tile.

If this is the intent, I think this is the first time we've done this in the codebase, and it probably makes sense to define a rectangle var with coords for each sub-image and apply it here so we don't rely on magic numbers for each.

OTOH, it may be better to stick with convention and have each small cross on a distinct tile.

(If it's just one image, change name to singular.)

mrimer (changes requested):

mrimer (approved):
Thank you!

PR #1172, opened 2026-08-28
RPG: Add functionality to create log of game actions from save

Hypexion:
This PR adds a new system to log game actions, and to output the result to a text file. This has been requested a couple of times.

This is done through a system of tracking "game events", which inherit from `CGameEvent`. Each class represents a thing that can happen, and holds the defining information for that event. They also have a function that creates a text description of the event. This was chosen over using cue events as many cue events lack all the details required, and some situations require tracking events across turns.

The logs are created using game logger objects, which inherit from `CBaseGameLogger`. This base class isn't abstract, but instead acts as a do-nothing implementation. There is currently only one implemented logger, `CStandardGameLogger`, which logs most game actions that affect a player's stats or equipment. Most "DROD actions" aren't logged, but a more comprehensive logger could extend `CStandardGameLogger` to do so.

Internally, `CStandardGameLogger` keeps a vector of game events that have been logged, which are converted to text and output to a buffer when the `output` method is called. The buffer can then be written to a given file using `writeToFile`. Currently, events are output whenever a player changes room. The standard logger only records events for a room if at least one event other than entering the room happened in that room.

By default, `CCurrentGame` creates a `CBaseGameLogger`, which is stored in a `unique_ptr`. `CCurrentGame::ActivateLogging` will create a standard logger, which will begin recording events.

Players can create a log from the Restore screen, as an new option from "Export". To create the log, we use `CDb::ValidateMoveSequence` to replay a save, activating the logger beforehand, then writing the log once the replay is finished. This only happens if an output file has been specified, so it should not impact CaravelNet verification. Log creation is fairly quick, although longer saves do take a noticeable time to process.

A few things have been refactored down into `DRODLib` to allow access. Additionally, `DrodFileDialogWidget` has been extended to support selecting `.txt` files. There's also a fix to a regression in here than I noticed while testing.

Thread: https://forum.caravelgames.com/viewtopic.php?TopicID=47462


drodrpg/DRODLib/CurrentGame.cpp
```diff
@@ -5409,9 +5465,12 @@ bool CCurrentGame::KnockOnDoor(CCueEvents& CueEvents, const UINT wX, const UINT
 			{
 				--ps.yellowKeys;
 				CueEvents.Add(CID_ItemUsed, new CMoveCoord(wX, wY, YellowKey), true);
+				logger->openDoorWithKey(YellowKey, wX, wY);
 			} else if (ps.skeletonKeys) {
 				if (!SpendSkeletonKey(CueEvents, wX, wY, ps)) {
 					return false;
+				} else {
```
mrimer: Style note: the "else" is superfluous on each of these skeleton key checks.

mrimer (approved):

drodrpg/DRODLib/CurrentGame.cpp
Hypexion: I think using else here makes the intend more clear, as it's marking the block as being deliberately entered. Although it might be cleaner to switch the logic around to something like:

```
if (SpendSkeletonKey(...)) {
  logger->doLogging(...)
} else {
  return false;
}
```

mrimer (approved):

